PR #14, opened 2020-12-21
tech team practices update

choldgraf:
closes https://github.com/2i2c-org/meta/issues/89


tech-team-practices.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Bi-weekly team syncs
+
+Every two weeks we synchronize our team goals and practices. Here's how to conduct a team sync.
+
+1. **Create an issue in `handbook/`**. This is where we'll keep track of the "to-do" for the team sync.
+2. **Clean up [the HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw?both) for the sync**. We'll use [this HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw?both) to run the team sync. At the top is a set of questions to answer, and we can We'll initially use the HackMD to give responses. You can create it from the following template:
+3. **Team members give their responses**. You can copy/paste the template, and then give your responses below. You shouldn't feel forced to add content if you can't think of anything, use it as much as is useful. Just remember that others depend on this information to know what one another is up to!
```
yuvipanda, line 30:
```suggestion
3. **Team members give their responses**. You can copy/paste the template, and then give your responses below. You shouldn't feel forced to add content if you can't think of anything, use it as much as is useful. 
```

The extra sentence seemed a bit too much 'stick' :)

tech-team-practices.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Per-project to-dos and conversation**: Generally speaking specific projects will have their own place for team conversation, to-do items, and strategy. This is generally *not* in the 2i2c handbook, but there should always be links and references in the handbook that points somebody in the right direction.
+
+**Discussion around team practices**: For discussion around strategy that doesn't have a natural place elsewhere, we use `handbook` issues and PRs. If an issue, it should be actionable (and "closeable"), generally in a way that results in a change to the handbook. If a PR, it should be a proposal to change the handbook and thus update some strategic information in there.
+
+## Bi-weekly team syncs
+
+Every two weeks we synchronize our team goals and practices. Here's how to conduct a team sync.
+
+1. **Create an issue in `handbook/`**. This is where we'll keep track of the "to-do" for the team sync.
```
yuvipanda, line 28: We could make this an issue template. Not a blocker for this though.

tech-team-practices.md
```diff
@@ -0,0 +1,40 @@
...
+2. **Clean up [the HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw?both) for the sync**. We'll use [this HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw?both) to run the team sync. At the top is a set of questions to answer, and we can We'll initially use the HackMD to give responses. You can create it from the following template:
+3. **Team members give their responses**. You can copy/paste the template, and then give your responses below. You shouldn't feel forced to add content if you can't think of anything, use it as much as is useful. Just remember that others depend on this information to know what one another is up to!
+4. **We'll leave the HackMD open for 48hrs**. That should give enough time for those of us across many time zones to response.
```
yuvipanda, line 31:
```suggestion
4. **We'll leave the HackMD open for 48hrs**, not counting weekends or holidays. That should give enough time for those of us across many time zones to respond.
```

Explicitly say that the 48h doesn't count holidays & weekends. 'Holidays' across countries is complicated but we can probably figure that out if it becomes a problem.

tech-team-practices.md
```diff
@@ -0,0 +1,40 @@
+# Staff sync practices
```
yuvipanda, line 1:
```suggestion
# Tech team internal coordination process
```

We should communicate that:

1. This is for the tech team
2. Once we have volunteers, will they also participate?
3. The goal of this is to co-ordinate between the teams, not to 'report up'.

I'm not sold on my suggestion either, would love to iterate.

choldgraf:
@yuvipanda I've updated this an applied a few changes per some of your comments. What do you think?

also - when do you want to officially start following this process? I think we should commit to doing it on the schedule we agree upon. Properly following these async practices will mean we need to be disciplined about informing one another what we are up to (I think that's the big challenge with doing things fully async, meetings *force* people to get on the same page by talking to one another, so if folks are regularly not filling out their team sync updates, it will be as if they just didn't show up to a team-wide meeting without telling people).

yuvipanda:
Let's give this a go :)

PR #16, opened 2020-12-24
renaming handbook to team compass

choldgraf:
After some discussion with @yuvipanda , I think it would be useful to rename this from "handbook" to "team compass". This more closely follows the practices we already follow in the Jupyter repositories, and which has worked well. This PR renames references to `handbook/` to instead use "team compass"

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -17,7 +17,7 @@
 
 # -- Project information -----------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


yuvipanda:
I like! Matches the jupyterhub and jupyterlab repos.

PR #17, opened 2020-12-29
adding a leads and sales page

choldgraf:
This adds process and references for generating sales and leads.

### DIFF
--- a/sales.md
+++ b/sales.md
@@ -1,0 +1,70 @@
+# The leads and sales process
+
+2i2c pursues new collaborations and contracts via following-up on leads for those who may be interested in 2i2c services.
+This page has information to help guide this process.
+
+(sales:questions-to-answer)=
+## Questions to answer
+
+To decide whether we should continue engaging with a particular lead, we need to know whether they are in-scope for our mission and likely to be able to successfully collaborate with us.
+The [`MANIAC-T` framework](https://xxiibrands.com/sales/qualify-your-sales-leads-with-maniac-t/) can be helpful in coming to a decision. Here is a short description of these guideilnes:
+
+* **Money**. Is there a specific amount of money set aside for the project? If they are seeking funding in order to pay for this, what are the details?
+* **Authority**. Does the contact have the authority to make a buying decision? Who else will need to be involved in the decision-making process?
+* **Need**. What are their clear pain-points with their current infrastructure setup? Do they seem to have a real need for change?
+* **Impending Event**. Do they have an upcoming event that is urgent or pressing?
+* **Application**. Is their use-case a good fit for 2i2c infrastructure? Would their use-case require major changes to our hubs? Is there a different organization better-suited to help them?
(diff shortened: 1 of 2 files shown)


sales.md
```diff
@@ -0,0 +1,70 @@
...
+We gather the most structured and useful information for our leads using [the leads generation questionnaire](https://docs.google.com/forms/d/1KHw-4Wdyoofv-6CENeiOSbnC62LKdx55QTcZmXQVkkc/edit?usp=sharing). This should have questions that allow us to gather the information noted in [](sales:questions-to-answer).
+
+As a part of the triage process, we'll send a lead a link to this questionnaire. Their responses should be linked in the GitHub issue for that lead.
+
+## Checklist for new leads
+
+Once someone or a team is interested in 2i2c, we should do the following things:
+
+1. Set up a short meeting (or email if they prefer) to understand their general use-case and needs. If they seem like they're in-scope for 2i2c, proceed with the next steps.
+2. Create a new GitHub issue called {guilabel}`<lead-name> - <short-title>` and add the following content to it:
```
yuvipanda, line 48: Let's make this an issue template instead?

yuvipanda (changes requested):
Great start! I made a couple of comments, otherwise LGTM

sales.md
choldgraf: Yeah - I was planning to get this merge in first so we can agree on the language, and then make an issue template about it in a follow-up. If you think the language is generally non-controversial then I'm happy to just make the template w/o creating a discussion phase.

sales.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Each lead gets an issue
+
+We use a GitHub issue to track the progress of each lead (created when we are reasonably sure the lead is worth following-up on). 
+We track leads in issues within the [`meta/` GitHub repository](https://github.com/2i2c-org/meta), using the {guilabel}`lead` tag.
+We keep a high-level view of all leads [at this project board](https://github.com/2i2c-org/meta/projects/2).
+
+### Each lead gets a folder
+
+We create a folder for each lead / customer. These exist in the `sales/` folder of the 2i2c Google Drive.
```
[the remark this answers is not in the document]
choldgraf: I think we should assume that any reference to a file/folder should be available to anyone with a 2i2c.org account by default, unless there's a reason for some should _not_ be accessible (e.g. maybe some hiring docs or something). In this case, I think it should be available to the team while we are at this size. Perhaps if 2i2c grows bigger then it will make sense to be more thoughtful about protecting some information

sales.md
yuvipanda: Ah, great. I think the template looks great for a start :)

sales.md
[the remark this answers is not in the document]
yuvipanda: Yep, I agree! let's record that here or somewhere else?

yuvipanda (approved):

choldgraf:
gonna merge this in so that I can start using it in our leads generation

PR #23, opened 2021-01-16
moving hub sales+infra tracking to a dedicated space

choldgraf:
This proposed change to our team practice will move all issues related to **sales + lead generation** as well as **running specific hubs** into the [`hubs/` repository](https://github.com/2i2c-org/hubs). The goal of doing this is to:

1. Further streamline the information clutter in `meta/` so that it's easier to track for team members
2. Have a dedicated source of truth for the hubs we run, so that we don't accidentally miss something
3. Have a dedicated source of truth for sales and leads, so that we don't accidentally miss something

### DIFF
--- a/sales.md
+++ b/sales.md
@@ -1,4 +1,4 @@
-# The leads and sales process
+# Sales and running hubs process
 
 2i2c pursues new collaborations and contracts via following-up on leads for those who may be interested in 2i2c services.
 This page has information to help guide this process.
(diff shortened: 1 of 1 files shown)


yuvipanda:
Specific issues about hubs should probably go into their own repos (like 2i2c-org/pilot-hubs or utoronto-2i2c/jupyterhub-deploy), but linked to from the project board in the 'hubs' repo.

choldgraf:
@yuvipanda yep that's what I imagined too. The `hubs/` repository wouldn't be a place to have discussion etc about individual hubs, it'd mostly be a collection of *pointers* to the hub-specific conversation that was happening elsewhere.

choldgraf:
OK I've updated this process to reflect the changes that we made in https://github.com/2i2c-org/pilot-hubs/pull/236 so I'm gonna merge this as I think it is now reflective of what we're doing!

Issue #32, opened 2021-02-15
Tech Team Update: 2021-02-15


choldgraf:
Hey @2i2c-org/tech-team - time to fill in some updates about what you've been up to the last couple of weeks!
Can folks fill out the [HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw) with their own updates? ✨ 

- **Updates HackMD**: https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw
- **Team Sync history**: https://2i2c.org/team-compass/team/tech/sync/

# ToDo

- [x] Clean up the [HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw) for this update
- [x] Ping the team members in [`#tech-updates`](https://2i2c.slack.com/archives/C01GLCC1VCN)
- [x] Wait 2-3 days
- [x] Copy/paste into the `team-compass` repository
- [x] Clean up the HackMD
- [x] Link to new updates in `team-compass/` in [`#tech-updates`](https://2i2c.slack.com/archives/C01GLCC1VCN)

PR #33, opened 2021-02-19
updating tech team coordination to include goals

choldgraf:
This updates our tech team process with information about how we keep track of team goals, and how this works with the team sync process.

Done in conjunction with https://github.com/2i2c-org/pilot-hubs/pull/236

cc @yuvipanda 

also updates with our latest team sync: closes https://github.com/2i2c-org/team-compass/issues/32


yuvipanda:
I like it! Thanks, @choldgraf

PR #38, opened 2021-03-08
Adding information about structure and governance.

choldgraf:
This PR defines our current governance and general structure of 2i2c, as well as our relationship to ICSI. It closes https://github.com/2i2c-org/meta/issues/133.

To re-iterate, the point here is not to create new governance, but to put names and labels to our current governing structures. For suggestions about *improvements* to our governance, see https://github.com/2i2c-org/meta/issues/158. But let's focus this PR around "describing" rather than changing.

I am not sure whether this PR requires a formal vote from the @2i2c-org/founders. On the one hand, it is a change related to governance, which in general I think requires votes. On the other hand, it is not trying to change the substance of our governance, only describe it explicitly.

If anybody believes that we need to have a founding team vote for this, please do not hesitate to suggest one. Otherwise, I'll plan to leave this open for several days to give people a chance to provide input.

You can find a preview of the page here: https://2i2c-team-compass--38.org.readthedocs.build/en/38/about/structure.html

closes https://github.com/2i2c-org/meta/issues/150
closes https://github.com/2i2c-org/meta/issues/133

### DIFF
--- a/about/structure.md
+++ b/about/structure.md
@@ -1,0 +1,60 @@
+# Structure and governance of 2i2c
+
+2i2c is an initiative dedicated to supporting open infrastructure in interactive computing for research and education, as well as the open source communities that underlie this infrastructure. Its host organization is {ref}`structure:icsi`.
+
+This page describes the major organizational structures of 2i2c, and how they relate to governance and operations.
+
+:::{admonition} A few notes on governance
+- All 2i2c members act as individuals, and not on behalf of their affiliated institutions.
+- All decisions made by the 2i2c Steering Council or Teams must abide by the policies of its host organization, {ref}`structure:icsi`.
+:::
+
+## Steering Council
+
+The Steering Council defines the mission, vision, and values of 2i2c. It also sets the strategic direction and priorities for 2i2c. The Steering Council provides oversight to the Executive Director of 2i2c and the Operations Team.
+
+The Steering Council makes decisions via consensus, and is restricted by the legal obligations and policies of ICSI. It uses the following process for these decisions:
+
+* Open an issue in the `meta/` GitHub repository using [the {guilabel}`decision` template](https://github.com/2i2c-org/meta/issues/new?assignees=&labels=decision&template=decision.md&title=%7B%7B+Decision+Summary+%7D%7D).
+* Have conversation around the decision in that issue. This may happen in the issue directly, or in ancillary issues, meetings, etc that are relevant. The decision issue should link to these relevant spaces.
+* Once everybody has voted, note the decision in the issue and close it.
+
+:::{seealso}
+The current Steering Council is [listed on the 2i2c website](https://2i2c.org/about/#steering-council).
+:::
+
+## Executive Director
+
+The Executive Director oversees the execution of the mission and strategy provided by the Steering Council.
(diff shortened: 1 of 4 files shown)


choldgraf:
Another ping for the @2i2c-org/founders on this issue. I'll leave it open until mid-next week. Unless anybody suggests that we need to have a vote on it, I'll plan to merge unless anybody suggests changes or has comments to discuss.

colliand:
This looks good to me. 👍🏽

about/structure.md
```diff
@@ -0,0 +1,60 @@
+# Structure and governance of 2i2c
+
+2i2c is an initiative dedicated to supporting open infrastructure in interactive computing for research and education, as well as the open source communities that underlie this infrastructure. Its host organization is {ref}`structure:icsi`.
```
rabernat, line 3:
```suggestion
2i2c is an organization dedicated to supporting open infrastructure in interactive computing for research and education, as well as the open source communities that underlie this infrastructure. Its host organization is {ref}`structure:icsi`.
```

about/structure.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Steering Council
+
+The Steering Council defines the mission, vision, and values of 2i2c. It also sets the strategic direction and priorities for 2i2c. The Steering Council provides oversight to the Executive Director of 2i2c and the Operations Team.
+
+The Steering Council makes decisions via consensus, and is restricted by the legal obligations and policies of ICSI. It uses the following process for these decisions:
+
+* Open an issue in the `meta/` GitHub repository using [the {guilabel}`decision` template](https://github.com/2i2c-org/meta/issues/new?assignees=&labels=decision&template=decision.md&title=%7B%7B+Decision+Summary+%7D%7D).
+* Have conversation around the decision in that issue. This may happen in the issue directly, or in ancillary issues, meetings, etc that are relevant. The decision issue should link to these relevant spaces.
+* Once everybody has voted, note the decision in the issue and close it.
```
rabernat, line 20: To my understanding, the notions of "voting" and "consensus" (above) are incompatible.

rabernat (commented):
Thanks for doing the hard work of thinking this through!

Overall LGTM!

about/structure.md
choldgraf: I was using initiative to prevent people from thinking we are a separate 501c3. Do you not share this concern?

about/structure.md
choldgraf: Ah - I was interpreting consensus to mean "every vote must be in favor in order to proceed". Is that not clear here? Or does it differ from your understanding of consensus?

rabernat: It was not clear to me.

about/structure.md
rabernat: The problem with "initiative" is that it sounds like it has a fixed time horizon. What happens when the initiative ends?

I see no problem with using "organization" in this way but don't feel strongly.

about/structure.md
choldgraf: I'll make that language more explicit 👍

choldgraf:
Updated the proposal per @rabernat's suggestions above 👍 

Will leave it open another day or two in case others have thoughts.

lheagy:
Thanks for pushing this forward @choldgraf! This looks good to me

choldgraf:
OK, I'm gonna merge this! Wohoo explicit instead of implicit!

PR #42, opened 2021-03-10
Adding updates to tech sync process

choldgraf:
This updates our team sync process in the following ways:

- Weekly instead of bi-weekly
- Uses updated github issue template instead of HackMD
- Updates the process describing this
- Adds a GitHub Action to open the weekly check-in on a CRON job

cc @GeorgianaElena @yuvipanda @consideRatio in case they have thoughts on whether this is a good approach!

### DIFF
--- a/.github/ISSUE_TEMPLATE/team-update.md
+++ b/.github/ISSUE_TEMPLATE/team-update.md
@@ -5,7 +5,7 @@ labels: team-sync
 title: "Team Sync - {{ date | date('dddd, MMMM Do') }}"
 ---
 
-This is a @2i2c-org/tech-team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team.
+This is a @2i2c-org/tech-team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team. This issue will be closed in 24 hours.
 
 ### Team Goals
 
--- a/.github/ISSUE_TEMPLATE/team-update.md
+++ b/.github/ISSUE_TEMPLATE/team-update.md
@@ -15,9 +15,7 @@ See the [Team Goal label](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Ai
 
 These issues require attention quickly:
 
-- **New Hubs**: The [Needs Hub issues](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Needs+Hub%22) are new hubs that need deployments.
-- **Needs Triage**: The [Needs Triage issues](https://github.com/2i2c-org/pilot/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Aneeds%3Atriage) require an initial assessment and labeling.
-- **Priority Issues**: The [Priority Label issues](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Apriority) are important and should be given attention over other issues.
+{GENERATE PROGRAMMATICALLY OR ADD HERE BY HAND}
 
 ### Team Updates
 
--- a/.github/ISSUE_TEMPLATE/team-update.md
+++ b/.github/ISSUE_TEMPLATE/team-update.md
@@ -2,7 +2,7 @@
 name: "🙌 Team sync"
 about: Syncronize the team's goals and actions
 labels: team-sync
-title: "Team Sync - [DATE]"
+title: "Team Sync - {{ date | date('dddd, MMMM Do') }}"
 ---
 
 This is a 2i2c Tech Team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team.
--- a/.github/ISSUE_TEMPLATE/team-update.md
+++ b/.github/ISSUE_TEMPLATE/team-update.md
@@ -5,7 +5,7 @@ labels: team-sync
 title: "Team Sync - {{ date | date('dddd, MMMM Do') }}"
 ---
 
-This is a 2i2c Tech Team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team.
+This is a @2i2c-org/tech-team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team.
 
 ### Team Goals
 
@@ -30,7 +30,7 @@ _Copy and paste these questions below, and answer them as you wish!_
 - So-and-so helped me out a lot with XXX...
 - Thanks for Jo's work on the XXX repo...
 
-**Updates from last two weeks ✔ **
+**Updates from last week ✔ **
 - I worked towards goal: <link-to-goal>
 - I merged issues XYZ
 - I had a meeting with ABC
--- a/.github/ISSUE_TEMPLATE/team-update.md
+++ b/.github/ISSUE_TEMPLATE/team-update.md
@@ -1,21 +1,47 @@
 ---
-name: Team updates
-about: Giving a team update
-title: "Tech Team Update: [DATE]"
-labels: team-process
+name: "🙌 Team sync"
+about: Syncronize the team's goals and actions
+labels: team-sync
+title: "Team Sync - [DATE]"
 ---
 
-Hey @2i2c-org/tech-team - time to fill in some updates about what you've been up to the last couple of weeks!
-Can folks fill out the [HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw) with their own updates? ✨ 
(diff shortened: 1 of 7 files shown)


.github/ISSUE_TEMPLATE/team-update.md
```diff
@@ -1,21 +1,47 @@
...
-- [ ] Ping the team members in [`#tech-updates`](https://2i2c.slack.com/archives/C01GLCC1VCN)
-- [ ] Wait 2-3 days
-- [ ] Copy/paste into the `team-compass` repository
-- [ ] Clean up the HackMD
-- [ ] Link to new updates in `team-compass/` in [`#tech-updates`](https://2i2c.slack.com/archives/C01GLCC1VCN)
+### Issues Requiring attention
+
+These issues require attention quickly:
+
+- **New Hubs**: The [Needs Hub issues](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Needs+Hub%22) are new hubs that need deployments.
```
yuvipanda, line 18: I like the specificity here

.github/workflows/team-sync.yaml
```diff
@@ -0,0 +1,16 @@
+# From https://github.com/JasonEtco/create-an-issue
+name: Create an issue on push
```
yuvipanda, line 2: This should probably have a different name

.github/ISSUE_TEMPLATE/team-update.md
```diff
@@ -1,21 +1,47 @@
...
-labels: team-process
+name: "🙌 Team sync"
+about: Syncronize the team's goals and actions
+labels: team-sync
+title: "Team Sync - {{ date | date('dddd, MMMM Do') }}"
 ---
 
-Hey @2i2c-org/tech-team - time to fill in some updates about what you've been up to the last couple of weeks!
-Can folks fill out the [HackMD](https://hackmd.io/i2Siurp1TkmPYgn3ZgxFQw) with their own updates? ✨ 
+This is a 2i2c Tech Team sync 🎉🎉🎉! This is a way for 2i2c Team Members to triage work, point out issues that require attention, and make sure we are working well as a team.
```
yuvipanda, line 8: Can we 'ping' the tech team here somehow automatically as well so it shows up in github notifications?

choldgraf:
thanks for the comments @yuvipanda - I've pushed a new commit with fixes for them!

yuvipanda:
Should we add a closing time requirement?

choldgraf:
@yuvipanda I imagined we will close the issue at the end of the working day California time (since that's beyond the end of the day for everyone else). Does that make sense?

I wonder if we could add another CRON job to do that...

yuvipanda:
@choldgraf would love another cron job doing that :) I'd also suggest midnight california time.

choldgraf:
@yuvipanda I *think* that the latest push adds auto-closing :-)

.github/ISSUE_TEMPLATE/team-update.md
```diff
@@ -1,21 +1,47 @@
...
-- [ ] Clean up the HackMD
-- [ ] Link to new updates in `team-compass/` in [`#tech-updates`](https://2i2c.slack.com/archives/C01GLCC1VCN)
+### Issues Requiring attention
+
+These issues require attention quickly:
+
+- **New Hubs**: The [Needs Hub issues](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Needs+Hub%22) are new hubs that need deployments.
+- **Needs Triage**: The [Needs Triage issues](https://github.com/2i2c-org/pilot/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Aneeds%3Atriage) require an initial assessment and labeling.
+- **Priority Issues**: The [Priority Label issues](https://github.com/2i2c-org/pilot-hubs/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Apriority) are important and should be given attention over other issues.
+
```
damianavila, line 21: Hi everyone! @choldgraf shared this PR with me and I am audacious (I am using the proper word??) enough to make some comments knowing just a little bit about this whole process :wink:

@choldgraf, I think you could be more date-specific in those links??
AFAIK, GitHub searching machinery is powerful enough to filter by dates.
If you can build these links to filter by date, then you will have a "permanent" record of the things happening on some specific week. Otherwise, if I am understanding this correctly, those links are always "live". And if you check the issue in some far away (or not that far away) future, those links (actually the content) will not be relevant for the specific anymore, I think...

.github/ISSUE_TEMPLATE/team-update.md
```diff
@@ -1,21 +1,47 @@
...
+**What I'm up to next ⬜**
+- I'd like to focus on goal <link-to-goal>
+- Developing on ABC feature
+- Focusing on XXX hubs
+
+**Links to items for discussion 💬**
+- I'm have a question about goal <link-to-goal>
+- Can @XXX give a comment on issue #NN ?
+- I opened #NN for discussion, please chime in
+```
```
damianavila, line 47: I like the discussion phase towards the end.
Otherwise, you end up discussing stuff for the whole meeting and you miss the important things before that phase (thanks, update, next).

team/tech/coordination.md
```diff
@@ -2,6 +2,22 @@
 
 As a totally distributed team, it is important to have team practices that give one another insight into what we're working on, identify opportunities to collaborate, and un-block one another. This page describes some of the practices that we adopt towards these goals.
 
+
+(coordination:team-syncs)=
+## Weekly team syncs
+
+Every Monday, the 2i2c Hub Operations team conducts a team sync to get on the same page and coordinate their work for the week.
```
damianavila, line 9: Weekly cadence is good IMHO.
Two weeks is actually a long time for sync, more than one iteration per week is noisy and, most of the time, it does not add enough value.

team/tech/coordination.md
```diff
@@ -2,6 +2,22 @@
...
+
+1. **Create an issue using [the team sync issue template](https://github.com/2i2c-org/team-compass/issues/new?assignees=&labels=team-sync&template=team-sync.md)**. This issue is our space to discuss, update, and sync.
+2. **Team members give their responses**. You can copy/paste the template, and then give your responses in comments to the issue. You shouldn't feel forced to add content if you can't think of anything, use it as much as is useful.
```
damianavila, line 12: > You shouldn't feel forced to add content if you can't think of anything, use it as much as is useful.

That's super important! People should not feel coerced to fill the blanks. With time, it should be natural and stress-free to share stuff here...

team/tech/coordination.md
```diff
@@ -2,6 +2,22 @@
...
+3. **Discuss and agree on next steps for 48hrs**. The goal is to ensure that important issues have somebody paying attention to them, and that team members are supported in the goals they work towards. After 48 hours, we'll close the issue.
```
damianavila, line 13: If you are talking here about the issue you opened for the sync, accordingly to the cron jobs listed above, I think you are actually closing it after 24+8 hs (I could be wrong, of course, since we are dealing with time shifts :wink:).
Since you automated the closing of the issue, I guess that block of time is only to agree on the management of the urgent matter, right? The alternative is to keep it open until the urgent matter is managed AND resolved...

choldgraf: Good point - fixed with slightly more vague and thus more correct language :-)

.github/ISSUE_TEMPLATE/team-update.md
choldgraf: ah that's clever! Lemme see if we can programmatically do that via the github action template

choldgraf: hmmmm now that I think about it, this might not be possible because GitHub attaches metadata to *issue events*, whereas what you're talking about is metadata associated with *when the search is being made*. AKA we might have 5 issues with `priority` open this week, but they may have all been *last updated* last week, and so we can make a query that says "give me a list of priority issues that were last updated between X and Y time" but if we were doing that for *this week*, the returned list would be empty. Does that make sense? Maybe I am missing something about the search API?

damianavila: I think you are right!
I guess the issue here is how to get the picture for each week.
Maybe some GH action to pull that "picture" using the GitHub API?
Maybe is as simple as adding a point here to remember to copy over that data every week?
(I prefer the automation, but maybe I am overthinking this...)

choldgraf: We could certainly add something like a "closed this week" update, something like that.

oooh you just gave me an idea for a little hacking project. I wanted an excuse to learn more about https://ghapi.fast.ai/ and I wonder if we could use that to write a github action that:

1. Pulls all open issues w/ `priorty` label
2. Converts it into markdown
3. Posts the GitHub issue using that markdown content to list issues explicitly.

Actually, we could get even fancier in the future and build something like: https://predictablynoisy.com/jupyter-activity-snapshot/jupyter.html

but that is probably not a blocker for this PR 😅

choldgraf:
OK I've heard only positive things about this PR, and it needs a little bit of debugging that requires merging first. So I am gonna merge it, and we can try piloting on Monday, and start to iterate a bit and improve things from there. Thanks all for your advice!

yuvipanda:
Thanks a lot for your feedback, @damianavila!

PR #50, opened 2021-03-13
improvements to team sync

choldgraf:
These are some debugs and improvements to the team sync process, so that more human-facing information is in the Issue Template and less in the Python script.

This also documents this process a bit better.

# To Do

- [x] For some reason I am getting an `Unauthorized` error when trying to read issues from a different team-compass repository. I'm not sure what's the right auth incantation to use here, so need to figure that out.


choldgraf:
Think I figured it out (which is code for got help from @hamelsmu and he is a better programmer than I am!)

PR #69, opened 2021-04-16
Updating sales pipeline

choldgraf:
Updates to our sales pipeline, per some recent conversations with Jacob.

### DIFF
--- a/managed-hubs/sales.md
+++ b/managed-hubs/sales.md
@@ -7,55 +7,85 @@ This page has information to help guide this process.
 
 Here are some locations of common documents used in the sales process (other than the repositories described below).
 
-- [The Sales folder in Google Drive](https://drive.google.com/drive/u/0/folders/1DEjtQ_hBw2Qae02_oe9UaGmRMYIkzJla) - contains all folders and information that cover the sales and contracting process.
-- [The leads folder in Google Drive](https://drive.google.com/drive/folders/1zBXL0X3S28W6iNAshBMaXF7gSpvnXa83?usp=sharing) - contains a folder for each lead that 2i2c is currently engaging.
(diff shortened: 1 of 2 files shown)


damianavila:
At overall this process feels OK to me.
Given that I was not part of the previous conversation I will defer the review approval for others.
One interesting point to discuss later is how to integrate this piece with the management process we are developing/trying to adopt:
* Should we generate a new project for each new Hub?
* Should we have a template project for that?
* How the prioritization is managed?
* Etc :wink:

choldgraf:
@damianavila for sure - I've opened up https://github.com/2i2c-org/pilot-hubs/issues/355 to track that issue, think we can merge this one and then continue updating in the future as we understand more what our team process should be?

damianavila:
> think we can merge this one and then continue updating in the future as we understand more what our team process should be?

Yep, I believe so...

damianavila (approved):

managed-hubs/sales.md
```diff
@@ -7,56 +7,86 @@ This page has information to help guide this process.
...
+
+Create a Lead Folder for them
+: As soon as a lead responds to the questionnaire, create a folder for them [in the leads folder](https://drive.google.com/drive/folders/1msuG0xWPbawMwesD8LZxdMClP-GVpgT2?usp=sharing). Use this to store any lead-specific information, such as price quotes and contracts.
+
+Decide if they are in-scope
+: If the lead is in-scope, then set the Triage status column to "In Scope" and move on to the negotiation stage.
+
+### Negotiation stage
+
+The negotiation stage is where we understand a bit more about what the lead needs and whether we'll be able to provide it. For example, do they want something really custom that goes beyond our standard Managed Hub distributions? If so, they might better as a "development lead" rather than a "managed service lead".
```
choldgraf, line 50:
```suggestion
The negotiation stage is where we understand a bit more about what the lead needs and whether we'll be able to provide it. For example, do they want something really custom that goes beyond our standard Managed Hub distributions? If so, they might be a "development lead" rather than a "managed service lead".
```

Issue #71, opened 2021-04-19
Improvements to the 2i2c Onboarding Documentation


choldgraf:
# Summary

We recently onboarded our first team member (using our current onboarding docs) and got a lot of useful feedback from this process. We should convert some of this feedback into improvements for the documentation! Summaries are below:

# Missing information

These were a few key pieces of missing information that were noted:

- [x] **How to navigate the team compass itself**. Much information was documented somewhere, but it wasn't clear how to look in the right place.
  - [x] Create an "introduction to the team compass" page that helps people navigate its contents
  - [x] Create/improve a "welcome to 2i2c" page for new team members that gives people an idea for the whole process
- [x] **Current team activity**. It wasn't easy to know what people were working on _right now_ (both at a high- and low-level).
  - [x] Document [our team workflow](https://github.com/2i2c-org/team-compass/pull/73), and how people can use it to understand this
- [ ] **What steps needed to be taken to onboard?**. We had a checklist in our team compass, but it wasn't clear who was supposed to carry out this checklist, and how to accomplish some of the items on there.
  - [ ] Complete / document https://github.com/2i2c-org/meta/issues/69#issue-732238413
  - [x] Include information in the team compass about how team members should provide access to these accounts
  - [x] Update [the onboarding checklist](https://team-compass.2i2c.org/en/latest/get-started.html#onboard-checklist-for-new-members) with new steps + links to how to accomplish them

- [x] **Describing 2i2c's overall structure and efforts**. It was unclear what 2i2c's "current organizational commitments" were. We have "projects", and "pilot hubs", but also had some that weren't mentioned like the Toronto + Farallon hubs.
   - [x] Clarify our active projects, where their repositories / information are located, as well as our hub information.
- [x] **List of all our currently-running hubs**. It was hard to find an up-to-date list of our currently-running hubs. [Our list of currently running hubs](https://pilot.2i2c.org/en/latest/about/projects.html#a-list-of-current-hubs) was out-of-date.
  - [x] Create a canonical list of all hubs that we run (https://github.com/2i2c-org/meta/issues/175)
- [x] **Define a welcome wagon**. 
  - [x] Send out a tweet about the person
  - [x] Add process for an "onboarding champion", who's job it is to carry out the onboarding process for a given person.

choldgraf:
@damianavila let me know if I'm missing anything on this one, these were the major points that I gathered from our conversations.

choldgraf:
[...]

PR #74, opened 2021-04-22
Updating onboarding and project information

choldgraf:
This is a refresh / addition to our onboarding and projects documentation. It takes a crack at several of the items here: https://github.com/2i2c-org/team-compass/issues/71

- Adds documentation about first places to look in the team compass (for those getting started)
- Adds an onboarding checklist as a github issue template and updates it
- Adds process about using the checklist + assigning an onboarding champion
- Adds pointers to the coordination page in the `get started` docs
- Updates our projects page with more recent information, as well as a list of hub infrastructure we currently run
- Updates some of the credentials etc that we need to create (though I suspect we may need to update this further)

Note: this adds a page with two tables representing our current hub deployment repositories, and the list of community hubs being served from `pilot-hubs`/: https://2i2c-team-compass--74.org.readthedocs.build/en/74/reference/hubs.html

If we like the list here, we should remove it from `pilot/` (though, I can see an argument for putting the latter list in `pilot-hubs/` rather than in team-compass)


choldgraf:
@damianavila asking for some 👀 from you since you are the most recent person to experience onboarding pain :-)

.github/ISSUE_TEMPLATE/new-team-member.md
```diff
@@ -0,0 +1,41 @@
...
+Follow the checklist below to onboard a new team member!
+
+**Communication**
+
+- [ ] Send them a link to the [Getting started docs](https://team-compass.2i2c.org/en/latest/get-started.html)
+- [ ] Schedule an onboarding meeting
+
+**Accounts**
+
+- [ ] Make email account
```
damianavila, line 21:
```suggestion
- [ ] Make email account
- [ ] Give access to the 2i2c Google Drive folder
```

.github/ISSUE_TEMPLATE/new-team-member.md
```diff
@@ -0,0 +1,41 @@
...
+- [ ] Add to [the proper GitHub team](https://github.com/orgs/2i2c-org/teams/2i2c-team)
+- [ ] Add to [the proper Slack group](https://2i2c.slack.com/admin/user_groups)
+
+**Permissions**
+
+- [ ] Add to appropriate Cloud projects
```
damianavila, line 27:
```suggestion
- [ ] Add to appropriate Google Cloud projects
```

conf.py
```diff
@@ -75,3 +75,50 @@
...
+
+# Pull latest list of communities served by pilot-hubs/
+import requests
+from yaml import safe_load
+
+yaml_list = [
+    "https://raw.githubusercontent.com/2i2c-org/pilot-hubs/master/config/hubs/2i2c.cluster.yaml",
+    "https://raw.githubusercontent.com/2i2c-org/pilot-hubs/master/config/hubs/cloudbank.cluster.yaml",
+    "https://raw.githubusercontent.com/2i2c-org/pilot-hubs/master/config/hubs/paleohack.cluster.yaml",
+]
```
damianavila, line 99: Don't you want to fetch all the `.cluster.yaml` here? Otherwise, this one will get outdated when we have a new cluster.

get-started.md
```diff
@@ -2,50 +2,73 @@
...
+2. [Open an issue in the Team Compass](https://github.com/2i2c-org/team-compass/issues/new/choose). Choose the "Onboard a new team member" template.
+   This issue will track the onboarding process, and will provide steps to take in order to onboard our new team member.
 
+The following sections describe some things that should be met in order to onboard somebody new.
 
-### NameCheap Access
+## Accounts that should be created
 
-2i2c runs much of its cloud infrastructure at custom URLs created for the target institution. In order to ensure you have the ability to create new URLs, you should have access to the 2i2c NameCheap account. If not, ask for access in the Slack.
+- **A GMail address**. You can use this to access information in the 2i2c drive, as well as various online services that we use such as cloud engines. You can sign in to your email address at gmail.com (it behaves like a Google account).
```
damianavila, line 42: Access to the 2i2c GDrive should be mentioned here, IMHO.

get-started.md
```diff
@@ -2,50 +2,73 @@
...
+- **A Slack account**. In addition to GMail, 2i2c also uses Slack to have informal or synchronous communication.
 
+## Accounts you should have access to
 
-### Cloud Accounts
+- **Google Drive**. Your 2i2c email address will get you access to a Google Drive that contains documents, brainstorms, and other assets that aren't fit for the 2i2c Team Compass. You can log into the Google Drive by going to `drive.google.com` and adding a new account using your `@2i2c.org` address and password.
```
damianavila, line 47: Actually, you needed to give me access to the 2i2c folder, otherwise, the drive is empty for the onboardee.

.github/ISSUE_TEMPLATE/new-team-member.md
choldgraf: Ah I think that the reason that access had to be explicitly given was because the folder was under a non `@2i2c.org` account - I've moved that to the `@2i2c.org` account, which means that anybody w/ an `@2i2c.org` address should now be able to access it.

conf.py
choldgraf: Are there others that I'm missing?

choldgraf:
OK I believe that I addressed @damianavila's comments via docs changes + google drive changes. I:

- Changed the "current list of hubs" to use the github API instead of manually specifying the YML files (we'll need to update this again if/when the YML file structure changes)
- Updated the google drive language to make it clear that the folder needs to be added
- Added a note about Staff vs. collaborators

.github/ISSUE_TEMPLATE/new-team-member.md
damianavila: That's super!! Thanks!

conf.py
damianavila: > Ah yeah - maybe a better approach would be to use the github API to grab these files, rather than hard code the yaml files here.

Yep...

> Are there others that I'm missing?

I don't think so, just thinking into the future...

damianavila (approved):
I think this is ready now.

choldgraf:
wohoo! thanks for the review

PR #81, opened 2021-04-27
Updating team sync for new project boards

choldgraf:
This updates our team sync with our new team workflow.

### DIFF
--- a/scripts/post-team-sync.py
+++ b/scripts/post-team-sync.py
@@ -17,71 +17,65 @@
 # %%
 from ghapi.all import GhApi
 from IPython.display import Markdown
-from textwrap import dedent
 from datetime import date
 from ghapi.actions import github_token
 import os
-
-# On GitHub Actions "ACCESS_TOKEN" should be a personal access token with r/w permissions to *other* repos
-token = github_token() if "ACCESS_TOKEN" not in os.environ else os.environ["ACCESS_TOKEN"]
-
-# %%
-# Initialize the GH API and our markdown
-api = GhApi(token=token)
-
-# %% [markdown]
-# # Base template
-# The base template is defined here: https://github.com/2i2c-org/team-compass/blob/main/.github/ISSUE_TEMPLATE/team-update.md
-# It has placeholders for lists of issues, and these will be automatically filled in below, then a new issue will be created.
-
-# %%
-# Grab our issue template
-from base64 import b64decode
-template = api.repos.get_content("2i2c-org", "team-compass", ".github/ISSUE_TEMPLATE/team-update.md")
-template = b64decode(template.content).decode("utf-8")
-
-# This removes the header bracketed by ---
-template = "---".join(template.split("---")[2:]).strip()
-
-# %% [markdown]
-# # New Hubs
-
-# %%
-issues = api.issues.list_for_repo("2i2c-org", "pilot-hubs", labels="Needs Hub")
-if issues:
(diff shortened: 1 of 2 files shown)


scripts/post-team-sync.py
```diff
@@ -17,71 +17,65 @@
...
-# Grab our issue template
-from base64 import b64decode
-template = api.repos.get_content("2i2c-org", "team-compass", ".github/ISSUE_TEMPLATE/team-update.md")
-template = b64decode(template.content).decode("utf-8")
-
-# This removes the header bracketed by ---
-template = "---".join(template.split("---")[2:]).strip()
+# Grab Ready to Work cards that point to issues of high priority
+boards = api.projects.list_for_org("2i2c-org")
+priority_labels = ['prio: high', 'priority']
```
damianavila, line 38: We should eventually use the same label in all the repos we own...

scripts/post-team-sync.py
```diff
@@ -17,71 +17,65 @@
...
+
+priority_issues = []
+for board in boards:
+    if "Activity" in board.name:
```
damianavila, line 42: So, I guess you are thinking here the Activity board content is too much fine-grained to be on this update, right?

scripts/post-team-sync.py
```diff
@@ -17,71 +17,65 @@
...
+#
+# Update our team sync issue template with some priority issues.
 
 # %%
-issues = api.issues.list_for_repo("2i2c-org", "pilot-hubs", labels="Needs Hub")
-if issues:
-    new_hubs = "\n".join([f"* [{issue.title}]({issue.html_url})" for issue in issues])+ "\n\n"
+# Read in our issue template
+if '__file__' not in globals():
+    curdir = Path(os.getcwd())
```
damianavila, line 63: Are you running/testing this interactively, right?

scripts/post-team-sync.py
choldgraf: yeah I feel like this issue is more like "the kinds of things I plan to focus on" rather than "the specific thing I'm doing tomorrow". That make sense?

scripts/post-team-sync.py
choldgraf: agree - I just put `priority` in case I missed something :-)

scripts/post-team-sync.py
choldgraf: yep, that's what this block is for

scripts/post-team-sync.py
damianavila: Total sense to me.

scripts/post-team-sync.py
damianavila: Thanks for the clarification!

damianavila (approved):
LGTM, thanks for all the answers to my questions!

scripts/post-team-sync.py
```diff
@@ -17,71 +17,65 @@
...
-# # New Hubs
+# # Update the template
+#
+# Update our team sync issue template with some priority issues.
 
 # %%
-issues = api.issues.list_for_repo("2i2c-org", "pilot-hubs", labels="Needs Hub")
-if issues:
-    new_hubs = "\n".join([f"* [{issue.title}]({issue.html_url})" for issue in issues])+ "\n\n"
+# Read in our issue template
```
choldgraf, line 61:
```suggestion
# Read in our issue template (using `__file__` check to see if we are not running interactively)
```

choldgraf:
great! thanks @damianavila :-)

PR #94, opened 2021-05-03
adding github labels to team conventions

### LINKED ISSUE #92: Create a set of standard Issue Labels to use across our repositories
# Background

Now that we have some team coordination and a workflow defined via [the activity board](https://github.com/orgs/2i2c-org/projects/5?fullscreen=true) it'll be more important that we provide proper categorization / metadata for the labels we use in these issues. We should define a minimal set of labels to use across all repositories, so that we can standardize their use and meaning.

Here are a few topics that we might want to track:

- Issue priority (`prio: low`, `prio: med`, `prio: high`) with increasing "red" colors
- Issue type (`type: enhancement`, `type: bug`, `type: documentation` etc)
- Issue state (e.g., "needs info", "needs upstream", "wont fix")
- Issue tags (`🏷 some tag`) - the tags shouldn't need to be standardized across repos, but the structure should be

# Steps to complete this goal

- [ ] Propose a set of tags via a PR to the team compass
- [ ] Get team feedback on this and converge on a solution
- [ ] Make a PR to add these to our `.github/` repository
- [ ] Do a one-time update to *add* all of these labels to our repositories

choldgraf:
This is a proposal to adopt a standard set of issue labels across all of our repositories. It defines a few common categories and their meaning.

Note that the PR is not really strict about forcing repositories to use these labels (except for the `type:` label), but is mostly concerned with standardizing structure and meaning. In the future if this system makes sense, we can consider requiring more of them, but I think it's best to start out simple.

closes #92

### DIFF
--- a/practices/github-conventions.md
+++ b/practices/github-conventions.md
@@ -1,6 +1,6 @@
 # GitHub repository conventions
 
-This page describes team 
+This page describes the use of GitHub Issue labels across the repositories within the 2i2c organization. 
 
 ## Issue labels
 
--- a/practices/github-conventions.md
+++ b/practices/github-conventions.md
@@ -7,6 +7,11 @@ This page describes team
 There are a few issue labels that we use to provide key metadata in our issues.
 These are added to all `2i2c-org/` repositories, and share the same meaning across each.
 
+:::{note}
+- Only **Issue Type** is required for all issues.
+- Repositories may define their own labels in addition to the ones described here for a given category, unless otherwise noted.
+:::
+
 ### Issue Type
 
 **REQUIRED**
--- a/practices/github-conventions.md
+++ b/practices/github-conventions.md
@@ -1,6 +1,6 @@
 # GitHub repository conventions
 
-This page describes the use of GitHub Issue labels across the repositories within the 2i2c organization. 
+This page describes some common conventions and patterns that we follow in our GitHub repositories.
 
 ## Issue labels
 
@@ -24,7 +24,6 @@ There are a few issue types that are defined for all repositories:
 - {badge}`type: enhancement, badge-info`: an improvement to be made to the repository.
 - {badge}`type: bug, badge-info`: something that is not working or incorrect in the repository.
 - {badge}`type: task, badge-info`: an action to take that is not well-categorized by enhancement/bug.
-- {badge}`type: goal, badge-info`: a high-level goal or outcome that we should work towards, with a specific completion criteria.
 
 In addition, other repositories may use repository-specific types, with the caveat that **all issues must still only have one `type` label**.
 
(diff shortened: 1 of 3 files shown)


practices/github-conventions.md
```diff
@@ -0,0 +1,73 @@
...
+
+Issue type determines the kind of issue and implies what sorts of actions must be taken to close it.
+Issue types are mutually-exclusive - there may only be one issue type per issue.
+
+There are a few issue types that are defined for all repositories:
+
+- {badge}`type: enhancement, badge-info`: an improvement to be made to the repository.
+- {badge}`type: bug, badge-info`: something that is not working or incorrect in the repository.
+- {badge}`type: task, badge-info`: an action to take that is not well-categorized by enhancement/bug.
+- {badge}`type: goal, badge-info`: a high-level goal or outcome that we should work towards, with a specific completion criteria.
```
damianavila, line 27: What is the interrelationship between the goal label and the concept of deliverables we have in the board structure?

practices/github-conventions.md
```diff
@@ -0,0 +1,73 @@
...
+**OPTIONAL**
+
+Issue state describes the current state of an issue, and implies the next action that must be taken in order to close it.
+Issue states are mutually-exclusive - there should not be more than one state label for an issue.
+Issues without a state label are assumed to be ready for implementation.
+
+There are a few issue states that are defined for all repositories:
+
+- {badge}`state: needs-info, badge-secondary`: the issue requires more information from the author in order to move forward. It may be closed after a few days of non-response.
+- {badge}`state: needs-discussion, badge-secondary`: the issue requires discussion and scoping from the team to create an actionable path.
```
damianavila, line 69: This one also overlaps with the board column you already have for discussion.

choldgraf: Agreed - it is unclear to me whether we can treat the board add complementary or independent. What is your intuition? Don't duplicate information across the board and labels?

practices/github-conventions.md
choldgraf: In my mind, a goal is more abstract and will likely happen over longer periods of time. E.g., "improve our team communication practices"

However, I think we could certainly recreate the same information with a combination of "enhancement" plus tags. Maybe we should just remove it for now?

damianavila: I think it will cause confusion, so I would be +1 on removing it for now.

practices/github-conventions.md
damianavila: Yep, do not duplicate info! Because otherwise, some people would codify/push info into the board and others into the labels... Both systems need to be complementary and provide different information, IMHO.

choldgraf: I'm that case, it feels like the boards should encode "state"...maybe we should remove the "state" label for now? Or, clarify how this state is different from a boards state?

I know that bokeh uses "reso" rather than state...I'm not sure whether that warrants it's own dedicated label type in our case. What do you think?

damianavila: > maybe we should remove the "state" label for now?

Yep, that will simplify things, IMHO.

> I know that bokeh uses "reso" rather than state...I'm not sure whether that warrants it's own dedicated label type in our case. What do you think?

Let's start simple, two set of labels such as `type` and `prio` is enough for now.

practices/github-conventions.md
GeorgianaElena: Should we also remove the goal issue template from the central `.github` repo https://github.com/2i2c-org/.github/tree/main/ISSUE_TEMPLATE? Though, looking at the repo now, it looks like the `new-goal` issue template doesn't have any effect on the org repos. I believe we should have the `ISSUE_TEMPLATE` dir inside a `.github` one for it to work.

Also, since we want the issue type to be a required label, maybe we could have some simple issue templates for this kinds of issue types in the `.github` that automatically add the label.

GeorgianaElena (approved):
I like the idea of having a labeling convention :+1: But I think we should make label emojis mandatory :D

choldgraf:
@GeorgianaElena that's a great idea - I'll implement your suggestions in the `.github/` repo if/when this gets merged

practices/github-conventions.md
choldgraf: @GeorgianaElena good point, and agreed. Though I think that if we define custom issue templates in any other repository, they *overwrite* the `.github` ones (AKA, they aren't *added* to the `.github/` ones, they replace them all)

choldgraf:
OK - I've updated the PR according to feedback! I think that this is mostly ready to go, depending on how serious @GeorgianaElena was about having emojis in all of our labels :-)

e.g., we could do something like:

 
- Tags: `🏷️ tagname`  
- Type: `☀️ enhancement` / `☀️ bug` / `☀️ task`
- Priority: `❗❗ low` / `❗❗ med` / `❗❗ high`

I'm fine going with that, but unless everybody is an enthusiastic +1, I don't think we wanna block the PR on this :-)

damianavila (approved):
I think this one is ready now.

PR #96, opened 2021-05-04
Time off for the team

choldgraf:
This is a lightweight proposal for telling one another when we have expected time away or reduced time. Mostly, it is just "Use a shared team Google Calendar to tell others when you're away".

see https://github.com/2i2c-org/team-compass/discussions/82 for more discussion around this.


damianavila (approved):
LGTM

PR #103, opened 2021-05-15
Point to pilot-hubs repo for list of hubs

yuvipanda:
- U Toronto hub is listed in https://github.com/2i2c-org/pilot-hubs/pull/413
- The pangeo-hubs repo is defunct, hasn't been actively used in a
  while and will be eliminated soon


yuvipanda:
Where do we track deliverables for the UToronto hub? Right now, nowhere! 

IMO, we should consider moving that hub into our pilot hubs repo as well. It was the first hub we setup, so it got its own repo. Now maybe we should move it back here. Will make life simpler for us.

index.md
```diff
@@ -56,7 +56,6 @@ Contains information about the 2i2c team and our projects, and some useful resou
 :maxdepth: 1
 :caption: Team Reference
 reference/projects
-reference/hubs
```
choldgraf, line 59: Imagined that we would leave this page, but just point it to the pilot hubs repo. That way the source of truth is still the team compass, it just delegates to pilot hubs. Maybe we can just add a section to the "team coordination" page?

yuvipanda: > Maybe we can just add a section to the "team coordination" page?

Can you expand on that a little?

choldgraf: I mean that if we think it is a bit much to have a dedicated page that simply links to the hubs table in this repo, we could turn that page into a short section here https://team-compass.2i2c.org/en/latest/practices/coordination.html

yuvipanda: @choldgraf that makes sense - let's do that!

choldgraf:
@yuvipanda is this ready to go?

damianavila:
I am not @yuvipanda 😉  but if we are removing the Hubs page here, we need to make an explicit link/reference to the table living in the pilots-hubs docs. @choldgraf you proposed to put the reference on the coordination page... I think it actually belongs to the https://team-compass.2i2c.org/en/latest/managed-hubs/about.html page, IMHO.

choldgraf:
Good point @damianavila - I've added a reference to that page in our `about` page that you linked above. Also resolved the conflict on `index` so I think we're ready to go unless anybody objects

damianavila:
@choldgraf, I do not actually see the reference you said you added when I look at the docs generated from this PR: https://2i2c-team-compass--103.org.readthedocs.build/en/103/managed-hubs/about.html.
Btw, the last commit is also a little bit weird... are you sure you pushed the change you were intended to push?

choldgraf:
Ermmm yep you're right, will look into it when I'm back at phone

choldgraf:
OK updated the PR! It didn't push properly when I tried earlier, thanks for catching it @damianavila :-)

damianavila (approved):
LGTM now.

choldgraf:
🚀

PR #108, opened 2021-05-25
Adding action to sort issues

choldgraf:
This adds a little GitHub Action that does the following:

- Loops through each column of our organizational Project Boards
- For cards that point directly to GitHub Issues, grab the labels for each issue.
- For issues with a "priority" label (e.g. `prio: high`), *sort the issues* based on high- to low-priority
- Sort that board's column so that the high priority cards are at the top

This should make it easier for us to keep the high priority issues at high-visibility for the team! It runs each day on a CRON job

(note I also ran `black` on the scripts so some minor formatting has changed)

cc @damianavila who might think this is neat :-)

### DIFF
--- a/scripts/post-team-sync.py
+++ b/scripts/post-team-sync.py
@@ -28,7 +28,9 @@ from pathlib import Path
 
 # %%
 # On GitHub Actions "ACCESS_TOKEN" should be a personal access token with r/w permissions to *other* repos
-token = github_token() if "ACCESS_TOKEN" not in os.environ else os.environ["ACCESS_TOKEN"]
+token = (
+    github_token() if "ACCESS_TOKEN" not in os.environ else os.environ["ACCESS_TOKEN"]
+)
 
 # Initialize the GH API and our markdown
 api = GhApi(token=token)
@@ -36,21 +38,24 @@ api = GhApi(token=token)
 # %%
 # Grab Ready to Work cards that point to issues of high priority
 boards = api.projects.list_for_org("2i2c-org")
-priority_labels = ['prio: high', 'priority']
+priority_labels = ["prio: high", "priority"]
 
 priority_issues = []
 for board in boards:
     if "Activity" in board.name:
         continue
-    columns = api.projects.list_columns(board['id'])
+    columns = api.projects.list_columns(board["id"])
     for column in columns:
-        if any(ii in column['name'].lower() for ii in ("in progress", "ready to work")):
-            for card in api.projects.list_cards(column['id']):
+        if any(ii in column["name"].lower() for ii in ("in progress", "ready to work")):
+            for card in api.projects.list_cards(column["id"]):
                 if "content_url" in card:
-                    _, org, repo, _, num = card.content_url.rsplit('/', 4)
+                    _, org, repo, _, num = card.content_url.rsplit("/", 4)
                     issue = api.issues.get(org, repo, num)
                     for prio_label in priority_labels:
-                        if any(prio_label in issue_label['name'] for issue_label in issue['labels']):
+                        if any(
+                            prio_label in issue_label["name"]
+                            for issue_label in issue["labels"]
+                        ):
                             priority_issues.append(issue)
 
 # %% [markdown]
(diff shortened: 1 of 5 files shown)


scripts/post-team-sync.py
```diff
@@ -96,6 +100,12 @@
...
 # %%
 # Now create an issue with this content
-resp = api.issues.create("2i2c-org", "team-compass", title=f"Team Sync - {date.today():%b %d, %Y}", body=template, labels=metadata['labels'])
+resp = api.issues.create(
+    "2i2c-org",
+    "team-compass",
+    title=f"Team Sync - {date.today():%b %d, %Y}",
+    body=template,
+    labels=metadata["labels"],
+)
```
damianavila, line 109: Using some new formatter? 😜

scripts/sort-issues.py
```diff
@@ -0,0 +1,86 @@
...
+#       extension: .py
+#       format_name: hydrogen
+#       format_version: '1.3'
+#       jupytext_version: 1.10.3
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
```
damianavila, line 15: ahhhh... prototyping in the notebook 😉

scripts/sort-issues.py
```diff
@@ -0,0 +1,86 @@
...
+                card["sort_label"] = label["name"]
+                cards_with_sorting_label.append(card)
+
+    # Sort the cards based on the sorting label
+    cards_sorted = sorted(
+        cards_with_sorting_label, key=lambda a: priority_list.index(a["sort_label"])
+    )
+
+    # Loop through the sorted cards (in reverse so we move the last label first) and move them to the top
+    for card in cards_sorted[::-1]:
```
damianavila, line 72:
```python
reversed(cards_sorted)
``` 
:wink:

scripts/sort-issues.py
```diff
@@ -0,0 +1,86 @@
...
+        print(f"\t\tMoved card {card['id']} to top")
+
+
+# Loop through our projects to sort cards
+for project in api.projects.list_for_org("2i2c-org"):
+    print(f"Sorting columns for project {project['name']}")
+
+    columns = api.projects.list_columns(project["id"])
+    for column in columns:
+        if any(ii in column["name"].lower() for ii in ["archive", "done"]):
```
damianavila, line 83: Is this over-complicated or am I missing something about the API response?
Is not enough to check if `column["name"].lower()` is in `["archive", "done"]` to `continue`?

damianavila:
> cc @damianavila who might think this is neat :-)

I think it is neat, indeed!
I have already provided some feedback.

scripts/post-team-sync.py
damianavila: And now I saw your comment about black :wink:

scripts/sort-issues.py
choldgraf: lol indeed that is a much more clear way of doing this 🤦‍♂️

scripts/sort-issues.py
choldgraf, line 72:
```suggestion
    for card in reversed(cards_sorted):
```

scripts/sort-issues.py
choldgraf: ah no nevermind, I knew there was a reason that I did it this way haha. The problem is that the columns aren't *just* named `archive` or `done`, so we just want to know if `archive` or `done` is _somewhere_ in the column name, but don't expect it to match the whole name

scripts/sort-issues.py
choldgraf: indeed - it means you can open up this python script *as a notebook* with jupytext, which is helpful for modifications and real-time prototyping

scripts/sort-issues.py
damianavila: OK, then a brief comment about it would be nice for the next reader.

scripts/sort-issues.py
```diff
@@ -0,0 +1,86 @@
...
```
damianavila, line 83:
```suggestion
        # columns aren't just named `archive` or `done`,
        # so we just want to know if `archive` or `done` is somewhere in the column name
        if any(ii in column["name"].lower() for ii in ["archive", "done"]):
```

choldgraf:
thanks for the comment addition @damianavila - sorry I missed it before! let's see if this works :-)

PR #120, opened 2021-06-14
Add RTD permissions to onboarding document

yuvipanda:


.github/ISSUE_TEMPLATE/new-team-member.md
```diff
@@ -32,6 +32,7 @@ Follow the checklist below to onboard a new team member!
 - [ ] Add to appropriate Azure projects
 - [ ] Add to the [auth0 2i2c tenant](https://manage.auth0.com/dashboard/us/2i2c/tenant/admins)
 - [ ] Add to [quay.io 2i2c team](https://quay.io/organization/2i2c/teams/owners)
+- [ ] Grant appropriate permissions on [RTD projects](https://readthedocs.org/)
```
consideRatio, line 35: What RTD projects are relevant to know about? It may be apparent if you are already onboarded and can see then listed though hmm...

I wonder if we should link specifically to certain projects or not, such as https://readthedocs.org/projects/oauthenticator/ for example. I guess we have a few projects though.

consideRatio: I think this may be good as it is since I can't figure out a way to make it clearer.

yuvipanda: Yeah I agree - can't figure out how to make it cleaner here.

PR #125, opened 2021-06-22
Rename remote to distributed

choldgraf:
This PR generally updates our team language to use the term "distributed" rather than "remote". As in, **2i2c is a distributed organization**, rather than **2i2c is a remote organization**.

Here are a few reasons for doing this:

- "remote" implies that there is a central home base from which others are remote
- "distributed" is more like an organizational strategy/philosophy ("our team is distributed"), whereas "remote" is often more about logistics ("I am working remotely today").

And [here is a nice post that discusses some of this](https://medium.com/walmartglobaltech/we-are-distributed-4c64ec6109f4).

What do folks think? Does this sound like a correct description of our team practices?

### DIFF
--- a/get-started.md
+++ b/get-started.md
@@ -67,13 +67,13 @@ You can use this to begin participating in our daily activities and find out wha
 
 Check out the [projects page](reference/projects.md) for more information about the projects that 2i2c is working on.
 
-## Tips for working remotely
+## Tips for distributed working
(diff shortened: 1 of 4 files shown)


damianavila (approved):
I totally agree with the change.
`Distributed` is what we should aim for, IMHO.

Issue #144, opened 2021-07-05
Team Meeting - July 6th


choldgraf:
We have our first @2i2c-org/2i2c-team meeting tomorrow! This is a reminder issue for all.

# Information

- **Date:** July 6th, 2021
- **Time:** 7AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/2021-07-06/07:00/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
- **Video conference link:** https://bit.ly/zoom-holdgraf
- **HackMD for meeting:** https://hackmd.io/Y5SBMxV7R6CMqzeTXgm5kA
- **Calendar for future meetings:** [this Google Calendar](https://calendar.google.com/calendar/embed?src=c_4hjjouojd8psql9i1a8nd1uff4%40group.calendar.google.com&ctz=America%2FLos_Angeles)

# Format

This meeting is partially an opportunity for us to speak with each other face-to-face a bit, and also a chance to synchronize and discuss important issues/plans/etc together. Since this meeting is quite new, let's discuss a bit what kind of structure works best for us.

# Actions

- [x] Have the meeting
- [x] Open up any issues or discussions as-needed to follow up
  - https://github.com/2i2c-org/team-compass/issues/148
  - Edits to https://github.com/2i2c-org/pilot-hubs/issues/355
  - Updates to https://github.com/2i2c-org/meta/issues/198
- [x] Cut notes out of HackMD and add to the Team Compass (https://github.com/2i2c-org/team-compass/pull/149) 
- [x] Update HackMD information for next month's meeting
- [x] Create a GitHub issue for next month's meeting: https://github.com/2i2c-org/team-compass/issues/163
- [x] Make modifications to this template as needed after the meeting
- [x] Turn this template into a GitHub issue template for future meetings
  - #160  
- [x] Meeting complete! 🎉

sgibson91:
Hmmm, I think the GitHub Action bot closed this a bit prematurely - I'll reopen

choldgraf:
I've updated the template above and turned it into an issue template for future meetings! (#160 #161 #162)

closing this one since we've got it all filled out :-)

PR #160, opened 2021-07-19
Team meeting template

choldgraf:
This adds a team meeting template for use in future team meetings!

related to #144


choldgraf:
gonna merge this one in since it's just a copy-over from #144 - we can update the template with different information over time!

PR #178, opened 2021-07-27
Update the roles section to be more complete

### LINKED ISSUE #166: Pilot Hubs should have a single non-2i2c admin user manually specified
# Summary

We have run into some confusion with setting admin users on some of our hubs, where people didn't know whether they were admins and were asking 2i2c for admin users status.

This is some extra noise that we have to deal with, and it would be simpler if there were a single, repeatable policy for who gets admin user status on a hub.

I propose that we adopt the following practice:

- All 2i2c engineers get admin user status on hubs via **manual configuration** (ideally this would be automatically interpolated from config somewhere)
- The Community Representative gets admin user status on hubs via **manual configuration**
- All other Hub Administrators must be added via the JupyterHub UI **by the Community Representative**.

# Important information

- Some discussion in slack: https://2i2c.slack.com/archives/C01DB2JRP8W/p1626768120154300
- An issue that led to this conversation https://github.com/2i2c-org/pilot-hubs/issues/529

# Tasks to complete

- [ ] Discuss and/or agree with this policy
- [ ] Write it up into the pilot hubs documentation
- [ ] Done!

Note: I think we can leave our *current* hubs as-is, and adopt this for any new hubs.

choldgraf:
This updates the "roles" page to have more complete information and to list the responsibilities of each role. It also embeds the diagram of our roles so that it is more discoverable.

closes #166 (along with https://github.com/2i2c-org/pilot-hubs/pull/557)


sgibson91 (approved):
👍🏻

Issue #190, opened 2021-08-02
Automate "Done" cards to move to the archive column after each deliverables sync meeting


choldgraf:

choldgraf:
closing this as it will be superceded by #188

PR #188, opened 2021-08-02
Update team meetings and deliverables workflow

choldgraf:
This PR updates our team workflows after a recent team meeting to discuss our Deliverables review meetings and how it fits in with board workflows. It also updates guidelines about meetings in general that we have recently discussed.

Major updates:

- Deliverables meeting is now weekly, with a noted plan to move back to bi-weekly
- Remove the dev/org backlogs from "official" team process, but includes description of them in case it's useful
- Simplifies language around the activity board and deliverables structure
- Updates our meeting templates with this new structure
- Encodes official meeting roles in the templates+docs

closes #201

PR #191, opened 2021-08-02
Add GHA to archive old cards

choldgraf:
This is a short script that uses `ghapi` to archive old cards in our Finished Card Archive. The goal of automating this is to ensure that:

1. The "Done" column reflects the work we've done in our most recent efforts (since it's useful to know what you've worked on over the last two weeks or so.
2. The "Archive" column has the cards from 4 weeks ago for reference, and afterwards those cards are cleaned from the board.

I wanted it to run every 2 weeks on Monday, but I couldn't figure out the CRON expression for this that wouldn't be super complicated 😬 

closes https://github.com/2i2c-org/team-compass/issues/190

### DIFF
--- a/.github/workflows/archive-old-cards.yaml
+++ b/.github/workflows/archive-old-cards.yaml
@@ -4,10 +4,7 @@ description: |
   are moved to an "archive" column, and cards from two cycles ago are
   "Archived" entirely (so they are not visible)
 on:
-  schedule:
-  # Run every 1st and 15th day of the month
-  - cron: "0 0 1,15 * *"
-
+  # We'll only run this manually since cleaning the Activity Board is part of the deliverables meeting. 
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/archive-old-cards.yaml
+++ b/.github/workflows/archive-old-cards.yaml
@@ -1,0 +1,33 @@
+name: Update Activity Archive
+description: |
+  This action updates the activity archive so that cards finished in the last cycle
+  are moved to an "archive" column, and cards from two cycles ago are
+  "Archived" entirely (so they are not visible)
+on:
+  schedule:
+  # Run every 1st and 15th day of the month
+  - cron: "0 0 1,15 * *"
+
+  workflow_dispatch:
+
+jobs:
+  open-sync-issue:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/archive-old-cards.yaml
```diff
@@ -0,0 +1,33 @@
+name: Update Activity Archive
+description: |
+  This action updates the activity archive so that cards finished in the last cycle
+  are moved to an "archive" column, and cards from two cycles ago are
+  "Archived" entirely (so they are not visible)
+on:
+  schedule:
+  # Run every 1st and 15th day of the month
+  - cron: "0 0 1,15 * *"
```
sgibson91, line 9:
```suggestion
on:
  schedule:
  # Run every 1st and 15th day of the month
  - cron: "0 0 1,15 * *"
  # Add button to manually run workflow
    workflow_dispatch: []
```

What do you think about adding a [manual trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) so we can also run this on the fly? In case a meeting is rescheduled or something?

choldgraf: Oh that's a cool idea! Actually what do you think if we took this one step further and:

1. Set this action to *only* be triggered by a manual click
2. Removed all the "check if two weeks old etc" logic from the action and just made it 1) Remove the cards from "Archive" and 2) Move "Done" to Archive
3. Added a step to our team deliverables sync meeting to "click the button to run this workflow"

sgibson91: I'm not sure about point 2. I can't decide if the logic helps or hinders the person running the workflow. But otherwise, yes :D

choldgraf:
OK the latest commit makes this a "push button" action, and adds a step to our deliverables meeting to manually run it! It also now uses Rich to make nicer logging messages.

Will need to be rebased likely once #188 is merged

damianavila:
This PR, I think, might be affected by @yuvipanda's proposal to have a per sprint board.
See comments [here](https://github.com/2i2c-org/team-compass/pull/193#discussion_r681753359) and [here](https://github.com/2i2c-org/team-compass/pull/188#discussion_r682401024).

choldgraf:
closing this as I believe it will be superceded by #188

Issue #163, opened 2021-07-19
Team Meeting: August 2021


choldgraf:
This is a planning issue for our next team meeting! See below for information about the meeting and next steps.

# Information

_Information about when and where the meeting will take place_

- **Date:** 2021-08-02
- **Time:** 7AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/2021-08-02/07:00/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
- **Video conference link:** https://bit.ly/zoom-holdgraf
- **HackMD for meeting:** https://hackmd.io/Y5SBMxV7R6CMqzeTXgm5kA
- **Calendar for future meetings:** [this Google Calendar](https://calendar.google.com/calendar/embed?src=c_4hjjouojd8psql9i1a8nd1uff4%40group.calendar.google.com&ctz=America%2FLos_Angeles)

# Follow up issues

_Links to follow up issues below_

- https://github.com/2i2c-org/team-compass/issues/192
- https://github.com/2i2c-org/team-compass/pull/188
- https://github.com/2i2c-org/team-compass/issues/167
- https://github.com/2i2c-org/team-compass/issues/175

# Actions

_Steps to take to have the meeting._

- [x] Have the meeting
- [x] Open up any follow-up issues or discussions and link above
- [x] Cut notes out of HackMD and add to the Team Compass: https://github.com/2i2c-org/team-compass/pull/193
- [x] Update HackMD information for next month's meeting
- [x] Create a GitHub issue for next month's meeting: #195 
- [x] Meeting complete! 🎉

choldgraf:
Hey @2i2c-org/tech-team - just a reminder that we've got a team meeting on Tuesday, so if you'd like to discuss something in particular please add it to the agenda!

sgibson91:
> we've got a team meeting on Tuesday

I think it's Monday

choldgraf:
Yes it's on Monday! 🤦

choldgraf:
Hey all - I've put some items on the team meeting agenda for today, mostly around discussing team process, encoding deliverables/goals in issues, etc. Please feel free to make any edits or additions if you think these aren't quite the right questions!

https://hackmd.io/Y5SBMxV7R6CMqzeTXgm5kA

choldgraf:
[...]

PR #193, opened 2021-08-03
Adding notes for August team meeting

choldgraf:
closes https://github.com/2i2c-org/team-compass/issues/163

### DIFF
--- a/meetings/eng/2021.md
+++ b/meetings/eng/2021.md
@@ -31,6 +31,7 @@ If you are joining the team video meeting, sign in below so we know who was here
         - The meeting's goal is to put things on the activity board.
         - Each item should have a person assigned to it
         - Should use best judgment about workload balance and making sure we're not overextended
+        - We should make a new board per each sprint, and treat it as immutable as possible
     - [x] De-commission the "Deliverables Backlog" and "Organizational Backlog" from "official board status"
     - [x] Simplify the "Activity Board" so that it has fewer columns and complexity.
         - Remove the "needs review" column, and instead write up a process to use GitHub review requests
--- a/meetings/eng/2021.md
+++ b/meetings/eng/2021.md
@@ -1,5 +1,130 @@
 # 2021 Meetings
 
+## 2021-07-02
+
+### Meeting roles
+
+- **Meeting facilitator**: @choldgraf 
+- **Meeting recorder**: @damianavila
+- **Meeting timekeepr**: @choldgraf
+
+### Welcome to the Team Meeting
+
+Hello!
+
+If you are joining the team video meeting, sign in below so we know who was here. Roll call:
+
+- name / GitHub handle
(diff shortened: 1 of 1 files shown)


meetings/eng/2021.md
```diff
@@ -1,5 +1,130 @@
...
+This is a place to make announcements (without a need for discussion), short updates about what you've been up to, and shout-outs to contributors! We'll read through these at the beginning of the meeting.
+
+- ICSI invoicing plan: https://github.com/orgs/2i2c-org/projects/5#card-65792797
+- We're going to get the CZI EOSS grant for Community Strategic Lead!
+    - This will open some addditional time/resources for pangeo stuff
+- Sarah is making progress on the pangeo deployment
+    - Some new terraform stuff
+    - This is some context on the headaches side https://github.com/2i2c-org/team-compass/issues/184
+        - Long process/negotation, when you say no, draw a line
+        - Private node is probably needed, but how we are going to manage the logings
```
sgibson91, line 53:
```suggestion
        - Private node is probably needed, but how we are going to manage the logins
```

typo

sgibson91 (approved):
LGTM!

meetings/eng/2021.md
```diff
@@ -1,5 +1,130 @@
...
+### Follow up items
+
+- [x] Chris sends the Binder cross stitch to his mom!
+- [x] Revisit times for the deliverable and monthly meetings (there are conflicts with other meetings)
+- [x] Define a strategy and high-level description of "the 2i2c pilot" https://github.com/2i2c-org/team-compass/issues/192
+- [ ] Updates to our board process https://github.com/2i2c-org/team-compass/pull/188
+    - [x] More tightly couple the "deliverables meeting" with the "activity board"
+        - The meeting's goal is to put things on the activity board.
+        - Each item should have a person assigned to it
+        - Should use best judgment about workload balance and making sure we're not overextended
```
yuvipanda, line 33:
```suggestion
        - Should use best judgment about workload balance and making sure we're not overextended
        - We should make a new board per each sprint, and treat it as immutable as possible
```

damianavila: If we are going into a sprinty way to manage the board, that makes sense...
Some worries about that is the proliferation of boards (unless you can somehow archive them and I think you can close them so that should be enough) and the fact the issues end tied to some specific board that could be useful from retrospective purposes but not that good to a wide project search (modulo that we have per project boards as we have discussed before).

choldgraf: IMO these notes are meant to be a reflection of the meeting's conversations and decisions, and I don't remember ever explicitly discussing whether we should use "one board per week, and treat it as immutable". Am I remembering incorrectly? I'm happy to discuss this but it feels like this makes it sound like we discussed and decided, when I don't believe we did.

PR #215, opened 2021-08-17
Add section on sops and secrets

choldgraf:
I was trying to debug the COESSING Grafana and realized I didn't have the password. I needed to get `sops` set up locally, and so used the opportunity to document the process a bit more thoroughly with some extra explanation.

This slightly duplicates content [in the pilot hubs docs](https://pilot-hubs.2i2c.org/en/latest/reference/tools.html?highlight=sops#sops) - perhaps we should cross link that one to here?

# Tasks

- [x] Consolidate and cross-link from other docs as well

### DIFF
--- a/practices/secrets.md
+++ b/practices/secrets.md
@@ -2,7 +2,7 @@
 
 This section describes how 2i2c team members share and access sensitive credentials that allow permissions on our infrastructure and services.
 
-## Location of information
+## Where are secrets located?
 
 **For more services**: Our team tries to use services that allow for multi-user authorization.
 This means we try to use services where team members have their own account with permissions on the service, so no secrets need be shared (for example, GitHub organizations).
@@ -10,6 +10,7 @@ This means we try to use services where team members have their own account with
 **When we must share secrets**: If we must share secrets for an account (for example, credentails for deployment to cloud infrastructure), then we use [the command-line tool `sops`](https://github.com/mozilla/sops) to encrypt our secrets.
 See below for information about how to use `sops`.
 
+(secrets:sops)=
 ## `sops` overview
 
 [`sops`](https://github.com/mozilla/sops) is a command-line tool for encrypting and decrypting secrets that are on disk.
@@ -43,8 +44,8 @@ As you can see, we have access to the **key names** but the **values are encrypt
 Some information about which encryption key `sops` uses to encrypt/decrypt the secret is also stored under the top-level `sops` JSON key.
 In the next section we'll cover how to unencrypt this file.
 
-(secrets:sops)=
-### Use `sops` to decrypt secrets
+(secrets:sops:setup)=
+### Set up `sops`
 
 To use `sops` with a 2i2c configuration file, follow these steps:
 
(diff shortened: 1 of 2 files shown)


practices/secrets.md
```diff
@@ -0,0 +1,62 @@
...
+3. **Set up the Google Cloud Key Management Service (KMS)**. This allows you to use your Google Cloud login to provide authentication for `sops`.
+   [Follow the `sops` instructions to use KMS](https://github.com/mozilla/sops/#encrypting-using-gcp-kms).
+4. **Decrypt the secrets file**. Once you've set up KMS, you should be able to decrypt the file with `sops`.
+   Run the following command, and you should see the secret values printed on your screen:
+
+   ```bash
+   sops --decrypt path/to/config/secrets.yaml
+   ```
+
+
```
sgibson91, line 62:
```suggestion
```{note}
Make sure to run `sops` from the root of the repository.
There are a set of rules stored in [`.sops.yaml`](https://github.com/2i2c-org/pilot-hubs/blob/master/.sops.yaml) that use regex to match a file to be encrypted with the encryption key location.
You will receive an error from `sops` if you are not in the root folder and it cannot see this file.
```
```

**Review by sgibson91 — APPROVED** [review]
Added a few extra points but this looks good!

**Comment by damianavila** [comment]
> This slightly duplicates content in the pilot hubs docs - perhaps we should cross link that one to here?

Yes, I think we need to consolidate the sops information in one place and link elsewhere...
There is also sop usage in other pilot-hubs pieces:

* https://pilot-hubs.2i2c.org/en/latest/howto/operate/manual-deploy.html (step 5)
* https://pilot-hubs.2i2c.org/en/latest/howto/operate/new-hub/add-aws-hub.html#create-the-cluster

File: practices/secrets.md [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+   Run the following command, and you should see the secret values printed on your screen:
+
+   ```bash
+   sops --decrypt --in-place path/to/config/secrets.yaml
+   ```
+
+```{note}
+Make sure to run `sops` from the root of the repository.
+There are a set of rules stored in [`.sops.yaml`](https://github.com/2i2c-org/pilot-hubs/blob/master/.sops.yaml) that use regex to match a file to be encrypted with the encryption key location.
+You will receive an error from `sops` if you are not in the root folder and it cannot see this file.
```
choldgraf, line 67:
````suggestion
```{note}
Make sure to run `sops` from the root of the repository.
There are a set of rules stored in [`.sops.yaml`](https://github.com/2i2c-org/pilot-hubs/blob/master/.sops.yaml) that use regex to match a file to be encrypted with the encryption key location.
You will receive an error from `sops` if you are not in the root folder and it cannot see this file.
```
````

**Comment by choldgraf** [comment]
OK I just merged in https://github.com/2i2c-org/pilot-hubs/pull/615#pullrequestreview-735218110 and since that + this both have approvals, and the content here hasn't changed much, I'm merging this in too! Thanks @sgibson91 for the review :-)

File: practices/secrets.md [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+   If you'd like to decrypt a configuration file in-place, use the `--in-place` or `-i` flag.
+   This will **overwrite** the configuration file with a decrypted version.
+   :::
+
+### Learn more about `sops`
+
+For more information about using `sops`, here are a few links to `sops` documentation:
+- [Basic usage of `sops`](https://github.com/mozilla/sops#usage)
+- [Encrypting files with GCP KMS](https://github.com/mozilla/sops#encrypting-using-gcp-kms)
+- [Common examples with `sops`](https://github.com/mozilla/sops#examples).a
```
damianavila, line 92: I think you have an additional `.a`, maybe you can remove it in the next PR @choldgraf 😉  ?

choldgraf: oops! I fixed this up in #220

PR #220, opened 2021-08-21
Add merge and review policy

choldgraf:
This is a proposed merge / review policy for our team's "non-infrastructure" changes! It also describes some of our decision-making process more generally!

For changes to active infrastructure, this PR refers to pilot hubs (and related PR: https://github.com/yuvipanda/pilot-hubs/pull/1)

### DIFF
--- a/practices/development.md
+++ b/practices/development.md
@@ -116,9 +116,10 @@ The Sprint Board is broken down into these columns:
 ## Merging and Reviewing policy
 
 The sections below describe major policies around merging and reviewing depending on the kind of change being made.
-Not all of them are followed strictly, though some are more important than others, and are marked with **REQUIRED**.
+There are some extra policies for changes that **affect actively-running infrastructure**.
+See the section below for details.
 
-### General Merge policy
+Not all of them are followed strictly, though some are more important than others, and are marked with **REQUIRED**.
 
 - **Always make a Pull Request**. (REQUIRED).
   All changes to 2i2c repositories should be made via a Pull Request.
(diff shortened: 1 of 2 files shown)


practices/development.md
```diff
@@ -112,12 +112,75 @@ The Sprint Board is broken down into these columns:
...
+- **Explicitly list a back-up if you must step away**. (REQUIRED)
+  PR authors are responsible for the infrastructure changes that they make.
+  You should generally only make a change to live infrastructure if you'll have the bandwidth to ensure it is fixed if something goes wrong.
+  If for some reason you **must** step away, make it clear who else is responsible for shepherding the PR.
+- **When you deploy leave a comment**. (REQUIRED)
+  You may deploy the infrastructure one or more times throughout a PR.
+  Each time you do, make sure the PR is updated with the state of the code when you deployed, and leave a comment making it clear to others that a deploy has been made.
+- **Use a draft PR before you have deployed something**.
+  This will signal to other members of the team that you have not changed any infrastructure yet, and are merely asking for feedback.
+- **Be more hesitant to deploy/merge without approval**.
```
yuvipanda, line 166: Most things we do in pilot-hubs are difficult to test without deploying, so I think we should relax this somewhat. I split this into two parts (new infra / existing) in https://github.com/2i2c-org/pilot-hubs/pull/516.

For most PRs, running it + seeing the response from that is going to be far far more important and useful than reviewing what's actually written. I suspect cases like https://github.com/2i2c-org/pilot-hubs/pull/613 will be common, where the code itself is incidental to the effects it has on deployment. We can't really wait for approval of the code before running it each time. 

I don't really have a suggested alternative tho.

practices/development.md
```diff
@@ -112,12 +112,75 @@ The Sprint Board is broken down into these columns:
...
+  For changes that are minor, very straightforward, and do not affect actively-running infrastructure, it is acceptable to self-merge a PR without getting an approval.  
+  If you don't believe that your PR requires an approval before merging, make it clear in your PR or in a comment that you plan to merge it in 24 hours.
+- **Leave PRs open for at least 24 working hours**.
+  This helps ensure that others on the team have have a chance to look at the PR and give their thoughts.
+
+In addition, there are special policies that apply for certain kinds of changes, described in the sections below.
+
+### Policy for terraform changes
+
+Changes that provision active hub infrastructure (e.g., almost anything involving `terraform`) have a different workflow that requires different policies.
```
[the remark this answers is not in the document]
choldgraf: ok cool - will re-work to include "all active JupyterHub changes"

practices/development.md
choldgraf: ah jeez I totally forgot that this PR exists...hmmm do you think it's better for us to move this section on "active infrastructure" to the `pilot-hubs/` repository and then cross-ref it from here? I could push to your PR if you like.

My reasoning for having it in `team-compass/` is that currently we do have a few non-`pilot-hubs/` repositories for infrastructure, but I could see that repository being a more natural place for people to look. What do you think?

damianavila: What I have written [here](https://github.com/2i2c-org/pilot-hubs/pull/618#issuecomment-902820512) is, I think, relevant for this discussion...

sgibson91: I did a bunch of work over in mybinder.org-deploy on deploying to a staging environment in a PR - most recently this PR which gets around the fact that if you've opened a PR from a fork, secrets are not available by default in GitHub Actions https://github.com/jupyterhub/mybinder.org-deploy/pull/2002

I'd be very happy to do something similar for the `pilot-hubs` repo

choldgraf:
ok merging this one in so that we can focus our attention on the meatier proposal in https://github.com/2i2c-org/pilot-hubs/pull/516

practices/development.md
sgibson91: Yes, to clarify, I was thinking on the hubs level with my above links, not on the cluster level which is now being discussed in #288 :)

PR #225, opened 2021-08-23
Add general issue template

choldgraf:
This adds a general issue template that has a good-enough structure to fit many cases, so that we don't have to choose from many different templates at once.


choldgraf:
merging this one in since we got two approvals in `pilot-hubs/` and I'm pretty sure the issue there is generic to all of our commonly-used repositories

PR #233, opened 2021-09-01
Checking links and fixing broken links

choldgraf:
final part of https://github.com/2i2c-org/pilot/issues/76


choldgraf:
the one broken link is only broken because we have not yet merged in https://github.com/2i2c-org/pilot-hubs/pull/516

Issue #227, opened 2021-08-25
Re-purpose our `prio` labels to be `impact` instead


choldgraf:
### Description

I propose that we remove the `prio:` GitHub labels and instead replace them with labels based on **impact** (so, `impact: low / med / high`).

We currently use a few GitHub labels to encode "priority": `prio:low / med / high`. I have found that these labels are under-specified. Priority is very context- and person-dependent, and also changes fairly frequently (e.g. what is low-priority yesterday may be high-priority tomorrow). Moreover, we currently have two different ways to encode priority: one is with the label, and the other is via an issue's placement in our development backlog / ordering on a column.

Instead I think we should use "impact" and define the meaning of this label specific to an issue's functional area. For example:

- the impact of an infrastructure feature might be defined by the number of users that would benefit from it (all users == high, ~half users == med, ~quarter users == low) OR particularly important users could trigger "high" as well.
- the impact of a bug might be defined by whether functionality is critically impaired or just cosmetic, or whether it is complex to reproduce or is reproduced by everyone
- the impact of an administrative task might be defined by whether it is required to get some essential work done, or just general housekeeping

There would still be a degree of subjectivity and judgment there, but I think this is still more concrete than "priority". In addition, the person that applies an "impact" label should ensure that the value/benefit is described in the issue well enough to justify the label.

### Value / benefit

If we instead used "impact" labels, I believe that it would be easier to have a concrete definition for each issue that would be more stable over time. Moreover, impact maps relatively nicely onto our "value/benefit/user story" practice in describing our team issues. Impact would be an important deciding factor in prioritization (in general, we want to work on things that are high-impact!) but it wouldn't **define** the priority, it would just be an important factor for it.

### Implementation details

I propose that we try this out on the `pilot-hubs/` repository for a month and see if it improves the signal-to-noise of the issues, or helps us triage and prioritize more easily.

If nobody objects, I'll plan to remove all `prio:` labels from pilot-hubs, and replace them with "impact" labels (and potentially re-assign labels to issues as needed).

### How to define "impact"?

Here is the definition we can use for impact across various kinds of categories:

- **Features / Enhancements**
  - `impact: high`: Will be seen and commonly used by nearly all users OR has been requested by an abnormally large number of users OR is of particular importance to a really important user
  - `impact: med`: Will be useful to many users but not an overwhelming amount. Most issues should be in this category.
  - `impact: low`: Is useful but not a critical part of workflows OR is a niche use-case that only a few users may need.
[...]

PR #252, opened 2021-09-27
Add impact issues

choldgraf:
Updates our documentation to include our use of `impact` labels. Also adds a noxfile to make it a bit easier to build the docs locally (I was running into a bunch of build errors because I have a bunch of dev versions installed in my global environment) 

Also updates `sphinx-panels` to `sphinx-design` since that's where the new development will be

I'll leave this open for a day or two, and if there are no suggested edits or objections I'll merge, since we've already discussed this a bit in the #227 

closes https://github.com/2i2c-org/team-compass/issues/227

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,31 @@
+# 2i2c Team Compass
+
+This repository contains the source files for the [2i2c Team Compass](https://team-compass.2i2c.org).
+For more information, see [our Team Compass landing page](https://team-compass.2i2c.org).
+
+## Build the team compass locally
+
+The Team Compass is built with [the Sphinx documentation engine](https://sphinx-doc.org).
+The easiest way to build the documentation in this repository is to use [the `nox` automation tool](https://nox.thea.codes/), a tool for quickly building environments and running commands within them.
+This ensures that your environment has all the dependencies needed to build the documentation.
+
+To do so, follow these steps:
+
+1. Install `nox`
+
+   ```console
+   $ pip install nox
+   ```
+2. Build the documentation:
+
+   ```console
+   $ nox -s docs_build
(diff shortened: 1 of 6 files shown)


choldgraf:
thanks for the approve @sgibson91 :-)

README.md
```diff
@@ -0,0 +1,31 @@
...
+   $ nox -s docs_build
+   ```
+
+This should create a local environment in a `.nox` folder, build the documentation (as specified in the `noxfile.py` configuration), and the output will be in `_build/html`.
+
+To build live documentation that updates when you update local files, run the following command:
+
+```console
+$ nox -s docs_live
+```
```
damianavila, line 31: @choldgraf, maybe we could have something like this for the pilot-hubs repo as well?

choldgraf: I would love that - the only challenge I ran into was that building the docs for pilot-hubs required installing an anaconda environment, and `nox` doesn't yet support native anaconda use. So it would be possible, but I think it'd mean that building the environment would be much slower because you'd have to manually install each time you built. I think there are some workarounds there though, I just haven't had time to try it out

damianavila: > nox doesn't yet support native anaconda use. 

I see `nox` in conda-forge: https://anaconda.org/conda-forge/nox, maybe I am missing something else?

damianavila: > maybe I am missing something else?

I guess the building time we are adding to the CI... do you know how much time that is?

choldgraf: yep - they do support conda commands directly, but not yet building a nox environment from a conda environment. The main dev does seem open to it though! https://github.com/theacodes/nox/issues/260

PR #254, opened 2021-09-30
Remove theme customizations and use 2i2c theme

choldgraf:
In a recent meeting with @colliand he provided helpful feedback that the documentation across 2i2c's pages was a bit confusing - some pages used the pydata theme, others used the sphinx book theme, and it was unclear how to get from one site to the other.

So, this PR demonstrates a new theme that I've created that we can use across all 2i2c documentation. It follows the same pattern that [the Dask documentation uses](https://docs.dask.org/en/latest/), which is to roughly do the following:

- Take a single-page theme (in this case, the `sphinx-book-theme`)
- Lightly modify it so that you add an extra topbar with hard-coded links
- Modify the theme to have a consistent style / branding

That way, every page of the docs has the same topbar, while the left navigation bar is unique to each docs. As a result, the experience of using a site with 2i2c is consistent across all of them. Over time, we might want to turn our topbar into drop-down menu items, but for now this is just a minimal set of links.

You can find the new theme here: https://github.com/2i2c-org/sphinx-2i2c-theme

but I thought that it would be easiest to demonstrate its usage via a "real world" PR.

Would love to know what people think. If folks are +1 on the idea, then I will update each of our documentation sites to use this theme.

another part of https://github.com/2i2c-org/meta/issues/78


choldgraf:
I've just pushed another update to the theme and rebuilt the preview here, after feedback from @colliand and @damianavila - would love to know what folks think. Here's a preview: https://2i2c-team-compass--254.org.readthedocs.build/en/254/

damianavila:
I think it is a great idea to have a unified experience with a common theme!

Some quick things to address in the future (no blockers for this PR):

Screen Shot 2021-10-01 at 10 11 56

1. Maybe all the links in the drop-down should be aligned to the right?
2. (I suspect it might be difficult) it would be nice that if you are in the team-compass, the drop-down "knows" that somehow and does not show the first item 😉  .

choldgraf:
agreed with both suggestions! They might be a bit tricky to get right (either in UI/UX or in implementation), so for now maybe we can merge this one in as an iterative improvement, and improve further from there!

PR #255, opened 2021-10-01
Add section on senior open source infra engineer

yuvipanda:
Companion PR to https://github.com/2i2c-org/2i2c-org.github.io/pull/68

Me and @choldgraf iterated on this a few times to bring myself
(and others!) clarity.

Co-authored-by: Chris Holdgraf <choldgraf@berkeley.edu>

### DIFF
--- a/positions.md
+++ b/positions.md
@@ -53,11 +53,6 @@ This section is primarily inspired by [On being a senior engineer](https://www.k
 
 - Act as a role model for positive, inclusive, and constructive team dynamics.
 
-### Director of engineering
-
-The director of engineering sets the technical strategy for 2i2c, as well as leads major technical initiatives for 2i2c infrastructure. Development work should be focused on architectural design and major infrastructure changes. This role defines the technical direction of 2i2c’s development efforts, and interfaces with 2i2c engineers to guide development in this direction.
-
-
 ### Director of 2i2c
 
 This role is the director of 2i2c. They oversee strategy, execution, and fundraising at 2i2c. They also ensure that 2i2c engineers are empowered to execute on their responsibilities, and thrive in their positions.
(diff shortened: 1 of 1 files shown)


yuvipanda:
We iterated on it here: https://hackmd.io/P36f_UEpRZ6kvvDAb0i-rQ?edit.

choldgraf (approved):
This looks good to me - thanks @yuvipanda for thinking this through. I'd love to know if others on the @2i2c-org/tech-team like the description here as well.


I'd also suggest we remove the "Director of Infrastructure" because I think that this role is under-specified right now. I think if we want it listed here, we should either spend time to think about its responsibilities etc, or we should just remove it, and since there's nobody in that role, I'm +1 on removing it :-)

yuvipanda:
@choldgraf removed!

choldgraf:
@yuvipanda I added a short intro to make it more similar to the other sections in structure, also turned your note into a footnote. What do you think?

yuvipanda:
I like the rewording! However, I think the link is important enough that it should be kept as a link and not as a footnote.

choldgraf:
Sounds good - can we put it after the bullet list? I am trying to think about the diataxis framework here. The "responsibilities" list is kind of like reference docs. We want it to be short and factual and to the point. Linking to material for inspiration is more like "explanatory" docs, so I think it should be separate from the reference part of the section rather than intertwined. Does that make sense?

choldgraf:
After a quick chat with @yuvipanda I moved the footnote back into the main description, since it is an important guiding document for the role. I didn't really change the content though. I'm +1 on merging this in.

yuvipanda:
Thanks for iterating on this with me, @choldgraf :)

PR #290, opened 2021-11-10
Update sprint planning meeting template

choldgraf:
Minor updates to our sprint planning meeting after some helpful brainstorming with @GeorgianaElena 

The major thing is that we've moved the "decide what to do with incomplete deliverables" section into the "planning what's next" section. That way we can focus the first part of the meeting just on recognizing what we've done.


choldgraf:
I'll merge this one now so that we can try it out for the next sprint meeting, but happy to iterate on the language in a new PR if there's more feedback.

PR #304, opened 2021-11-15
Add readthedocs config

choldgraf:
Explicitly adds a readthedocs configuration file, so that we can pin to Python 3.8 (and because this is a best-practice for RTD)


.readthedocs.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+
+python:
+  version: "3.8"
```
choldgraf, line 4:
```suggestion
  version: "3.8"  # To ensure that we can install the sphinx-2i2c-theme
```

.readthedocs.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+
+python:
+  version: "3.8"  # To ensure that we can install the sphinx-2i2c-theme
+  install:
+    - requirements: requirements.txt
+
```
consideRatio, line 7:
```suggestion
# Configuration on how ReadTheDocs (RTD) builds our documentation
# ref: https://readthedocs.org/projects/2i2c-team-compass/
# ref: https://docs.readthedocs.io/en/stable/config-file/v2.html
version: 2

build:
  os: ubuntu-20.04
  tools:
    python: "3.8"

sphinx:
  configuration: conf.py

# Explicitly opt out of trying to build additional formats such as PDF and ePub
formats: []

python:
  install:
    - requirements: requirements.txt
```

If you want to ever update to Python 3.9, you must use the non-deprecated way of installing Python, which is like this.

I added some misc stuff I think can be helpful to have declared in a RTD file as well by copy-pasting from other configs.

choldgraf:
hmm - the sphinx-2i2c-theme is now failing to build due to the enviroment missing `node`, but this was working with the old configuration style (https://readthedocs.org/projects/2i2c-team-compass/builds/15282403/)...any idea why that would have changed?

consideRatio:
Hmmm, I figure the build.os -> ubuntu:20.04 makes them use a different base environment?

.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
+# Configuration on how ReadTheDocs (RTD) builds our documentation
+# ref: https://readthedocs.org/projects/2i2c-team-compass/
+# ref: https://docs.readthedocs.io/en/stable/config-file/v2.html
+version: 2
+
+build:
+  os: ubuntu-20.04
+  tools:
+    # The sphinx-2i2c-theme requires python >= 3.8
+    python: "3.8"
```
consideRatio, line 10:
```suggestion
    python: "3.8"
# Set the version of Python and other tools you might need
build:
  os: ubuntu-20.04
  tools:
    python: "3.9"
    nodejs: "16"
```

choldgraf:
well the sphinx-theme-builder should automatically download a node environment to the home directory, which is why I'm confused

.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
...
+
+sphinx:
+  configuration: conf.py
+
+# Explicitly opt out of trying to build additional formats such as PDF and ePub
+formats: []
+
+python:
+  install:
+    - requirements: requirements.txt
```
consideRatio, line 20:
```suggestion
    - requirements: requirements.txt

```

choldgraf:
yep that seemed to be the issue. I've opened up https://github.com/pradyunsg/sphinx-theme-builder/issues/3 as a result but I think for now we can just define node in the config like you've done here

consideRatio (approved):

choldgraf:
Merging this since it un-breaks our docs builds!

PR #306, opened 2021-11-17
Open sprint issue every two weeks

choldgraf:
The goal of this PR is to make our "open a sprint planning meeting" issue to open an issue every **two weeks** instead of every week. It uses the following logic:

- The action runs at the same time each week.
- It defines a `start_date` that is the reference date that we'll use to define "every two weeks"
- It takes the current date, and figures out the number of weeks since the `start_date`.
- It takes the modulo 2 of that difference.
- It checks if the modulo is 0, and stores this as a `bool` as an environment variable.
- The "open issue" step then checks the value of that environment variable before deciding whether to move forward.

I'm not positive that this PR as it stands actually works, so would love any comments on whether this looks correct to you :-)

### DIFF
--- a/.github/workflows/create-sprint-planning-meeting.yaml
+++ b/.github/workflows/create-sprint-planning-meeting.yaml
@@ -15,8 +15,10 @@ jobs:
     - name: Calculate if we're on a 2-week cycle
       run: python datetime.py
       id: is-two-weeks
+    - name: Echo the weeks thus far 
+      run: echo "Is-two-weeks is ${{ steps.is-two-weeks.outputs.IS_TWO_WEEKS }}"
     - name: Open a new issue
-      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS }} == 'True'
+      if: ${{ steps.is-two-weeks.outputs.IS_TWO_WEEKS }} == 'True'
       uses: JasonEtco/create-an-issue@v2
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/create-sprint-planning-meeting.yaml
+++ b/.github/workflows/create-sprint-planning-meeting.yaml
@@ -16,7 +16,7 @@ jobs:
       run: python datetime.py
       id: is-two-weeks
     - name: Open a new issue
-      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS == 'True' }}
+      if: steps.is-two-weeks.IS_TWO_WEEKS == 'True'
       uses: JasonEtco/create-an-issue@v2
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/create-sprint-planning-meeting.yaml
+++ b/.github/workflows/create-sprint-planning-meeting.yaml
@@ -16,7 +16,7 @@ jobs:
       run: python datetime.py
       id: is-two-weeks
     - name: Open a new issue
-      if: steps.is-two-weeks.IS_TWO_WEEKS == 'True'
+      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS }} == 'True'
       uses: JasonEtco/create-an-issue@v2
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


sgibson91 (approved):
Couple of typos, but other than that, I say merge and see if it works! I'll be really interested to see if the print statement in Python works as expected 🙂

.github/workflows/create-sprint-planning-meeting.yaml
```diff
@@ -11,7 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: JasonEtco/create-an-issue@v2
+    - uses: actions/setup-python@v2
+    - name: Calculate if we're on a 2-week cycle
+      run: python datetime.py
+      id: is-two-weeks
+    - name: Open a new issue
+      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS == 'True' }}
```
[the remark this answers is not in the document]
sgibson91: This is a lie, sorry! (editing...)

.github/workflows/create-sprint-planning-meeting.yaml
```diff
@@ -11,7 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: JasonEtco/create-an-issue@v2
+    - uses: actions/setup-python@v2
+    - name: Calculate if we're on a 2-week cycle
+      run: python datetime.py
+      id: is-two-weeks
+    - name: Open a new issue
+      if: steps.is-two-weeks.IS_TWO_WEEKS == 'True'
```
sgibson91, line 19:
```suggestion
      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS }} == 'True'
```

Eep, my comment got mangled (on my phone, sorry!), this is the suggestion I meant to send 

Eep, my comment got ma

choldgraf:
I wonder if we could help debug a bit by adding an echo in there, so we know what the value of the environment variable is

sgibson91:
@choldgraf yes `echo ${{ steps.is-two-weeks.outputs.IS_TWO_WEEKS}}` should do that

.github/workflows/create-sprint-planning-meeting.yaml
```diff
@@ -11,7 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: JasonEtco/create-an-issue@v2
+    - uses: actions/setup-python@v2
+    - name: Calculate if we're on a 2-week cycle
+      run: python datetime.py
+      id: is-two-weeks
+    - name: Open a new issue
+      if: ${{ steps.is-two-weeks.IS_TWO_WEEKS }} == 'True'
```
sgibson91, line 19:
```suggestion
      if: ${{ steps.is-two-weeks.outputs.IS_TWO_WEEKS }} == 'True'
```

sgibson91:
Or ssh into the action with tmate https://github.com/mxschmitt/action-tmate

sgibson91:
> so we know what the value of the environment variable is

To be clear, we are setting an output, not an env var. So this will show up in the GitHub context, not the shell environment.

choldgraf:
ok I think the latest commit fixes those issues, I will merge this in and trigger a dispatch to see what happens :-)

choldgraf:
Note - I may do a little pushing-directly-to-main as I debug if there are tiny things to fix

choldgraf:
A few commits I added to main:

https://github.com/2i2c-org/team-compass/commit/0c4bdec954711e16d4fe8a2716a0c614035a83a4
https://github.com/2i2c-org/team-compass/commit/41f5c7fa529d84c8fc98db0a98f14db87afd0fc4
https://github.com/2i2c-org/team-compass/commit/69b20ebdbbc02e9eb24716bed3673ef197b429fe
https://github.com/2i2c-org/team-compass/commit/4f515dc7962d7cfdb2ce15d04beb8b0c1b4246fb
https://github.com/2i2c-org/team-compass/commit/7a7b187db0035a3ecd5b155f0df93c2b2bb66c1d
https://github.com/2i2c-org/team-compass/commit/c15112d1a78fa01fe5688a0a4927bd9bd668280e

(sorry, it is hard to debug GHA without just pushing to main and giving it a shot!)

PR #313, opened 2021-11-24
Fix action timing for sprint planning

choldgraf:
I think I figured out what was going wrong with our sprint planning action opener, so this PR:

- Calculates the number of days since the day *after* our last sprint planning meeting, that way we correctly detect whether we are nearing the end of week 1, or week 2 of a sprint
- Runs this every Tuesday instead of Wednesday, so that we have a day's notice reminder that the meeting is happening


choldgraf:
I'm gonna merge this one, then manually trigger the action to see if it would have correctly opened an issue today

Issue #312, opened 2021-11-24
Sprint Planning Meeting: 2021-11-24


sgibson91:
# 2i2c Sprint Planning meeting

This is a **60 minute** recurring meeting **every other Wednesday at 7:30AM Pacific time** (here's [**a timezone converter**](https://arewemeetingyet.com/Los%20Angeles/2000-01-01/07:30/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=), ignore the date!).

Our goal is to synchronize the team on the most important things to work on, and to divide work between one another so we know what to work on next.

- [**Sprint Board**](https://github.com/orgs/2i2c-org/projects/21)
- [**Deliverables backlog**](https://github.com/orgs/2i2c-org/projects/22)

### Meeting roles

- **Meeting facilitator**: @sgibson91 
- **Next Support Steward**: @sgibson91 

### Meeting agenda

_Follow the agenda below, checking boxes as we complete each step._

**Review the Sprint Board.** (10m)

- [x] Celebrate our accomplishments
- [x] Archive deliverables we finished

**Populate Sprint Board with new deliverables** (45m)

- [x] Review in-progress deliverables from the last sprint: drop or carry over items as needed
- [x] Review a few more urgent items in the [deliverables backlog](https://github.com/orgs/2i2c-org/projects/22/views/1)
- [x] Add new deliverables to Sprint Board
- [x] Somebody is assigned to each deliverable
- [x] Everybody agrees the current cycle's plan is realistic and that the right items are on the board.

**Support steward** (5m)

- [ ] Support steward describes any open support issues.
- [ ] Support steward transition issue is opened

GeorgianaElena:
@2i2c-org/tech-team, I forgot to ask about the next Support Steward and if we plan to do it in teams this time, or during the next round?

choldgraf:
After some conversation in Slack, here is our plan for support steward etc.

**Problem**: @sgibson91 is currently the meeting facilitator, but she is also next in line for support steward. So we need to increase the gap in time between "being meeting facilitator" and "being support steward". The easiest way to do this is to increase the number of people between these two roles in the rotation.

**Next steps**:

- We will skip @yuvipanda who was going to be the next meeting facilitator, to increase the gap in time between these roles. The next meeting facilitator will be @choldgraf
[...]

Issue #294, opened 2021-11-10
Team Roles Meta Issue


consideRatio:
### Description

This is a meta-issue to keep track of who is currently on the Support Team. When we transition support stewards, update the top comment of this issue with the latest information.

### Support Team

Our current support stewards are:

- @choldgraf 
- @yuvipanda

### Meeting Facilitator

Our current meeting facilitator is:

- @choldgraf

consideRatio:
Some takeaway points:
- Good thing to always reply quickly with a "I've received this ticket" or "I'll look into this" etc.
- Perhaps we need a procedure about how to go about closing tickets or similar?
- We have merge/split tickets techniques that can be used in freshdesk
- We have add private note as a technique that can be used to for example summarize a situation for a ticket, helping other people not need to re-read the entire tickets discussion.
- We also enabled email notifications on https://2i2c.freshdesk.com/a/admin/email_notifications

consideRatio:
If this issue is meant as an issue to be closed by "Tasks to transition the support steward!", absolutely we close it.

I figured it was more like an issue to keep open until the stewardship is over and the next one has opened, like the jupyterhub/team-compass typically has an issue open to represent the next team meeting.

choldgraf:
I think we should either:

- Leave this issue open to reflect who is the current support steward
- Have some other place where we keep track of who is the support steward (e.g., a google sheet that is also displayed on the team compass?) and then add a task to this issue that says to update that sheet.

Anyone have opinions?

consideRatio:
I'm opinionated towards leaving it open.

choldgraf:
let's just leave the issue open for now, and we can re-visit if there's a better practice for tracking this in the future

sgibson91:
> * Leave this issue open to reflect who is the current support steward

We could pin the issue too, so it's always at the top of the issues list?

choldgraf:
I'm +0, since I rarely look at the issues list . But if it would be useful for others I'm +1 on giving it a shot

sgibson91:
> since I rarely look at the issues list directly

It's the first place I go if I'm _searching for a specific issue_ since it's not guaranteed to be on the sprint or backlog boards. And I think "Who is the current support steward?" is a scenario where we are looking for a specific issue.

choldgraf:
I've repurposed this issue to serve as a "meta issue" that lists our current roles. I've also opened a PR to include a step to update this issue when those roles changed. Finally, I've pinned this issue so it has high visibility!

PR #318, opened 2021-11-24
Update roles and backlog practices from last meeting

choldgraf:
This makes a few minor team process updates after our last meeting:

- Adds some references to [our team roles meta issue](https://github.com/2i2c-org/team-compass/issues/294) which is the issue we can use to track who is in what role at each moment in time.
- Simplifies our support steward transition issue a bit, and reduces some of the redundant information
- Documents our practice of assigning people to an issue once it's actively being worked on, and defines our team's current "Work in Progress" as "the issues that have somebody assigned to them"
  - This is the biggest change in this PR, it means that we should take a strict approach to issue assignment, so that we *only* assign somebody when they're actively working on it, and if somebody is no longer actively working on an issue then we should remove their assignment from it.

red #312 and some conversations in there and after

### DIFF
--- a/.github/ISSUE_TEMPLATE/meeting-sprint-planning.md
+++ b/.github/ISSUE_TEMPLATE/meeting-sprint-planning.md
@@ -17,7 +17,6 @@ Our goal is to synchronize the team on the most important things to work on, and
 ### Meeting roles
 
 - **Meeting facilitator**: {{ NAME }}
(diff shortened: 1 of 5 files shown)


sgibson91 (approved):
LGTM!

damianavila (approved):
LGTM, I like the mapping between assignment and WIP... it is actually a really quick way to see the WIP on the backlog board.

PR #323, opened 2021-12-02
Merge support steward action with sprint planning action

choldgraf:
This merges the support steward transfer action with our sprint planning action, so they always happen at the same time.


damianavila (approved):
LGTM

PR #344, opened 2022-01-04
NEW: Team roles calendar

### LINKED ISSUE #294: Team Roles Meta Issue

### LINKED ISSUE #334: Create a "team roles calendar"
### Description

We currently use [a team roles meta-issue](https://github.com/2i2c-org/team-compass/issues/294) to track which team members are in each role. However, this is a bit clunky to update, and also doesn't make it easy to see when team members will take on a role in the *future*.

Instead, we can create a Google Calendar that tracks all of our team roles. We can update this instead of a GitHub issue	

### Value / benefit

A few benefits:

- Calendars are easier to track into the future
- We can compare this calendar with our other team calendars, most importantly the **Team Leave** calendar - this could help us spot when people are planning to be away while they are in an important role
- It feels more natural to track dates in a calendar, than in an issue, so it may be more findable.

### Implementation details

_No response_

### Tasks to complete

_No response_

### Updates

_No response_

choldgraf:
This adds a team roles calendar to keep track of when team members are in each role, rather than documenting this via #294 . I've created that calendar and populated it for the next 3 months with the team members that are serving in each role.

closes #294    
closes #334

### DIFF
--- a/projects/managed-hubs/roles.md
+++ b/projects/managed-hubs/roles.md
@@ -3,10 +3,6 @@
 There are a few major roles that we define for the Managed JupyterHub Service.
 These are outlined below.
 
-:::{seealso}
-They are also described in [the Hub Services diagrams slideshow](https://docs.google.com/presentation/d/1kqrviwVOoZfey_rujhIasdkKZmlYgxV-1J2AG-nr3VY/edit#slide=id.ge3f2127292_0_573) as well as in the [Managed Service Plan](https://docs.google.com/document/d/1Ka7tgJe7HR8EmS_MMakrYztgfkJT_iFksPsWHdQBqhM/edit?usp=sharing).
-:::
-
 (roles:community-representative)=
 ## Community Representative
 
@@ -56,32 +52,11 @@ People in these roles are generally affiliated with 2i2c.
 - Debug and resolve major issues with a hub that require intervention from a Hub Engineer
 - Perform open source development on technologies that are in use by the hubs
 
-## Support Steward
+(roles:support-steward)=
+## Support Stewards
 
-:::{warning}
-This is an experimental role and the details may change!
-:::
-
-The Support Steward is tasked with keeping track of ongoing support requests to `support@2i2c.org`.
-They do not necessarily complete the request themselves, and should work with other engineers to ensure they are resolved.
-We use [this GitHub issue to track which team members are currently serving in each role](https://github.com/2i2c-org/team-compass/issues/294).
-
-The Support Steward rotates throughout the engineering team each sprint, in order to ensure that all team members share the load of supporting our communities.
-
-See [the Support Process proposal](https://docs.google.com/document/d/17Kj_FbtVMl32TEcfvCp18fF1SEiBjVOhCswdidUytgM/edit?usp=sharing) for the latest version of our support process.
-
-```{button-link} https://github.com/2i2c-org/team-compass/issues/294
-:color: primary
-Team Roles membership issue
-```
-
-### Responsibilities
-
-- Check the communication channels for support that we are using
-- Respond to new support requests
-- Communicate with community representatives throughout the support process
-- Coordinate with engineering team members to ensure support requests are completed
-- Pay attention to ongoing support efforts to ensure they are resolved in a timely manner
+Support Stewards are the first point of contact for all support-related issues.
+See [](support:process) for more information.
 
 ## Diagram of roles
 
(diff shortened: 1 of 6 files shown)


projects/managed-hubs/roles.md
```diff
@@ -56,32 +52,11 @@ People in these roles are generally affiliated with 2i2c.
 - Debug and resolve major issues with a hub that require intervention from a Hub Engineer
 - Perform open source development on technologies that are in use by the hubs
 
-## Support Steward
```
choldgraf, line 59: This section is mostly duplicated in our dedicated support page now, so removing it

sgibson91 (approved):
Love it! 🙌🏻

sgibson91:
Can we link this new calendar to a Slack channel for the notifications? I'm not sure which would be best, `#team-updates`? I don't think there's a need for it to be private though?

GeorgianaElena (approved):
:rocket:

choldgraf:
I've set up a Slack integration to post a **weekly summary** for this calendar to the `#team-updates` channel. I'm not sure exactly what this behavior will be like in Slack, but let's give it a shot and see how it feels.

Will merge this one in soon unless there are objections!

damianavila:
@choldgraf, in the calendar, I noticed you are doing the switch for the Support role on Tuesdays, is that intended, right?
I had the impression the switch was happening on Wed in the previous iterations...

choldgraf:
I made the support steward position end on a Tuesday so that we didn't have a triple overlap in calendar events - so at 11:59pm on Tuesday it's person A, and at 12:01am on Wednesday, it's person B. So the switch does still technically happen on Wednesday. Does that make sense?

GeorgianaElena:
@choldgraf, I just noticed this on the calendar. Shouldn't the support role be passed over every two weeks?
support-roles


Also, it looks like you're both doing the support and meeting facilitator roles at the same which we agreed that should be avoided :(

sgibson91:
> so at 11:59pm on Tuesday it's person A, and at 12:01am on Wednesday, it's person B

This is timezone specific though. So from _my_ view, it looks like the handover happens Mon -> Tue, not Tue -> Wed. Might be similar for you too, @damianavila?

Screenshot 2022-01-06 at 09 59 44

damianavila:
> Might be similar for you too, @damianavila?

Maybe? I tried changing my timezone to PT (momentarily) and still see it on Tue > Wed

choldgraf:
Thanks for the feedback all - a few quick responses:

- @GeorgianaElena good point that I'm doubled up somehow. Hmm, what if we moved the **Meeting facilitator** role to you, and then started the rotation from there? This just means that you'd facilitate the next team meeting (not the sprints). What do you think?
- @GeorgianaElena re: "every two weeks", I thought we'd agreed that we will spend a month in this role, now that we have 2 support stewards at the same time, so they'd overlap by 2 weeks each. Did I remember that wrong?
- @sgibson91 re: "timezones" that's a good point - hmm, I don't think there's going to be any way to satisfy this for all timezones. Can we just treat these calendars as a general guide, and have a team understanding that "tuesday, in whatever timezone makes most sense for you, is the day that the role changes people"?

GeorgianaElena:
> You remember correctly, but somehow in Georgiana's screenshot, she is scheduled to become Support Steward only one week after Erik has taken on the role. So it looks like there's an error there?

+1 on what @sgibson91 said :+1:  I mistakenly said "the support role be passed over every two weeks". The correct way to put this into words I guess is "every two weeks a new person joins the support team".

> @GeorgianaElena good point that I'm doubled up somehow. Hmm, what if we moved the Meeting facilitator role to you, and then started the rotation from there? This just means that you'd facilitate the next team meeting (not the sprints). What do you think?

Hmm, I believe the next team meting is on Feb 1st and since I'm going to be support steward then, then it becomes an overlap for me. Instead of constantly switching the starting point of the round robin (we we just noticed it doesn't work), why not have a policy of equitably choosing a backup person or something like that when these things happen? Or some other form of knowing who is next in line?

choldgraf:
@GeorgianaElena ah that is a good point, damnit my brain does not process dates well haha.

OK let's skip it forward one more slot - @sgibson91 would you be able/willing to serve as the Meeting Facilitator for our next team meeting, and we can begin the rotation from there?

sgibson91:
I mean, yes, but then I'm due to become support steward on Feb 15th (I'm also thinking of taking time off then too) :sweat_smile: - perhaps this "not wearing both hats at the same time" rule is a bit too strict while we're still such a small team? :confused:

choldgraf:
@sgibson91 yeah, I have the same intuition that you do. Since the Meeting Facilitator now is only responsible for 1 meeting, rather than all of our bi-weekly meetings, perhaps it's not such a problem if there is partial overlap?

sgibson91:
Why is the meeting facilitator no longer responsible for the sprints?

damianavila:
I have not looked at this in detail but I think there will be always an overlap because the two cycles (support and facilitator) have different velocities.

And given that we are a small team, I do think as well we should not be that strict about the overlap...

> Yeah, I agree. The astronomer in me wants to say "UTC for everything! It has both 'universal' and 'coordinated' in the name!" - but that's maybe a step too far 😄

UTC, UTC, UTC 😜  (just kidding...)

Issue #259, opened 2021-10-04
Add a regular asynchronous team check-in


choldgraf:
### Description

Now that we've merged to two-week sprint cycles as part of https://github.com/2i2c-org/team-compass/issues/182, it might be helpful for us to have a lightweight way for us to check-in with one another throughout the week. In a team meeting, there was interest in trying out an asynchronous team check-in process. It would have the following goals:

- Be an opportunity to ask for feedback or input
- Help team members provide accountability and transparency to one another
- Identify opportunities to coordinate or hand-off work
- Identify whether there are unexpected challenges we didn't anticipate when scoping work
- Provide ourselves an easy way to look back and see what we've been up to this week/month/etc
- Help us focus on just one or two things at a time, and not spread ourselves across too-many daily projects

In addition, it'd have these constraints:

- Must be very lightweight - responding shouldn't require any "thought" and should only take 30-60 seconds
- Should be asynchronous, and not require people to respond at awkward times
- Should be followed by all team members (and should be light-weight enough that it's not a chore to maintain the practice)

### Value / benefit

Having a practice like this will make it easier for us to signal to one another what we're up to throughout the week, since we have less time to speak with one another face-to-face in our planning. Having a lightweight check-in process will help us accomplish the goals in the description above.

### Implementation details

In our meeting we had discussed something like a check-in every couple of days, but after thinking about it a bit, I think it might be the easiest if we just try to adopt something that is **daily** on weekdays. I think having it each day will make it more likely to just become a habit, and will also force us to keep the process lightweight enough that it does not feel like a chore to carry it out.

So I propose that we do the following:

- For the next 2 months, we try the following process and revisit in December 2021.
- **When**: Wednesdays and Fridays, but skip Wednesdays with sprint planning. AKA, every Friday, and every other Wednesday. At 5:00pm India time (or 5:00pm in each person's time zone if we can figure it out).
- **How**: Use a [Slack reminder](https://slack.com/slack-tips/customize-team-reminders) (or if it is too clunky to do this at the same time for all w/o changing time-zones, use the [Geekbot](https://geekbot.com/))
- **What**: A DM that is sent to each person, or a questionnaire that is posted in a channel, with these questions to answer:
    - How are you feeling today? ✨
    - What did you work on today? 💪
    - What are you up to next? 🏃
    - Would you like help with anything in particular? 🙏
 
  When you answer them via the DM, they'll get posted to the `#team-updates` channel

That's it!

[...]

PR #347, opened 2022-01-10
ENH: Convert team syncs from issues to slack only

choldgraf:
This converts our team sync process to use the Slack Geekbot instead of GitHub issues each week. Here are major changes:

- Archives our actions and issue template around creating a weekly team sync. I've decided to archive them rather than delete them for now, in case we want to re-visit this practice. But I've added notes to delete them in a few months if we don't change things up.
- Moves our weekly sync process into the development coordination guide, since I think that's a more natural place for it.

ref https://github.com/2i2c-org/team-compass/issues/259

### DIFF
--- a/practices/development.md
+++ b/practices/development.md
@@ -30,7 +30,7 @@ Wednesday (beginning of sprint)
 During the sprint
 : Team members work on the items assigned to them at the sprint planning meeting.
   We use [the Sprint Board](coordination:sprint-board) to coordinate our activities during the sprint.
-  On Mondays, each team member fills out a **team sync update** to share the major things they worked on, and note any unexpected challenges or blockers.
+  We provide updates about what we've been up to, what we're doing next, and where we need help via regular **asynchronous Slack stand-ups**.
 
 Tuesday of week 2 (end of sprint)
 : By the end of the day, team members should have completed all of their items for that sprint.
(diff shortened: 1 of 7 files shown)


practices/development.md
```diff
@@ -74,6 +74,32 @@ The Sprint Board is broken down into these columns:
...
+Throughout the week we have a lightweight asynchronous team sync process so that we can get on the same page.
+The primary goals of this process are:
+
+- To ensure that nobody is stuck on something
+- To signal-boost requests for review and help
+- To provide accountability of what we've been up to
+- To help us coordinate what to do next
+
+We use [the Geekbot](https://geekbot.com/) to manage this process.
+During a sync, each team member gets a message in their morning time with a few questions.
```
GeorgianaElena, line 90: I noticed that this Monday's bot message was at 7PM and not in the morning as usual.

GeorgianaElena (approved):
This looks good to me.
Two small things I noticed though:
- the bot doesn't send the messages in the morning
- should we give more details about the Slack updates? I noticed there isn't any info about the M-W-F workflow. But maybe this would make sense in the TODO section.

practices/development.md
choldgraf: I wonder if that was just because I flipped the switch at that time and it sent a backlog of messages? Let's see what happens on Wednesday

choldgraf:
@GeorgianaElena could you go into more detail on your second point? What extra information did you have in mind, and where? I'm happy to add in some extra detail somewhere!

damianavila (approved):
LGTM

choldgraf:
Another ping for @GeorgianaElena to provide some guidance on what you mean by "give more details about the Slack updates". I'm happy to make changes if you wish, just not sure where you'd like them to be! If you don't think it's a blocker or don't have time to look by tomorrow, I'll merge this one in and we can always iterate

practices/development.md
GeorgianaElena: Yes, the Wednesday message was at 10 AM :tada:

GeorgianaElena:
> should we give more details about the Slack updates? I noticed there isn't any info about the M-W-F workflow. But maybe this would make sense in the TODO section.

@choldgraf, sorry for my message being confusing :facepalm: When I said "more details" I was referring to mentioning that the updates happen M-W-F. Definitely not a must or a blocker.

Sorry for the confusion again!

practices/development.md
```diff
@@ -74,6 +74,32 @@ The Sprint Board is broken down into these columns:
 
 In addition, we have a few other pieces of metadata to signal different kind of actions that would be needed 
 
+
+(coordination:team-syncs)=
+## Daily team syncs
+
+Throughout the week we have a lightweight asynchronous team sync process so that we can get on the same page.
```
choldgraf, line 81:
```suggestion
Throughout the week we have a lightweight asynchronous team sync process so that we can get on the same page.
These currently happen on Monday, Wednesday, and Friday in the morning time of each team member.
```

choldgraf:
Ah thanks for that clarification @GeorgianaElena - I've added a sentence to convey this information as well! Will merge now :-)

PR #357, opened 2022-01-24
Document our styleguide/tools we use with pre-commit

sgibson91:
This adds a styleguide to the practices section of our team-compass documenting which tools we use with pre-commit

closes https://github.com/2i2c-org/infrastructure/issues/938


practices/styleguide.md
```diff
@@ -0,0 +1,13 @@
+# Styleguide
```
choldgraf, line 1:
```suggestion
# Style guide for repositories
```

Just to clarify a bit (also I _think_ style guide is two words but I am not positive haha)

choldgraf (approved):
Looks good to me, with one quick suggestion!

PR #360, opened 2022-01-31
Add Jan meeting notes

choldgraf:
This adds our meeting notes from January 2022.


sgibson91 (approved):

choldgraf:
I think the linkcheck is working on `main` now so merging. Will fix if it's legit broken

PR #374, opened 2022-02-22
Add some more AWS SSO instructions

yuvipanda:


damianavila (approved):
LGTM, in fact, I was able to self-serve a password reset whereas I was checking these steps.

yuvipanda:
Thanks, @damianavila!

PR #375, opened 2022-02-22
Add compensation reference docs

choldgraf:
In designing the new role for https://github.com/2i2c-org/meta/issues/256 , we used the GuideStar report (provided by CS&S) to have a more data-driven definition of salaries for similar organizations. I wrote up a short set of guidelines for the process that we followed to define the salary for this role, so that we are transparent about where our salaries come from and so we can re-use it in the future.


sgibson91 (approved):
LGTM

PR #376, opened 2022-03-01
Add a rationale to our compensation page

choldgraf:
As a part of https://github.com/2i2c-org/2i2c-org.github.io/pull/93 I'm updating our compensation page a bit. We had a long paragraph about our rationale there, but I think this makes more sense as a part of our compensation strategy documents, and we can cross-link it from our 2i2c.org site.

### DIFF
--- a/practices/secrets.md
+++ b/practices/secrets.md
@@ -10,13 +10,17 @@ This means we try to use services where team members have their own account with
 **When we must share secrets**: If we must share secrets for an account (for example, credentails for deployment to cloud infrastructure), then we use [the command-line tool `sops`](https://github.com/mozilla/sops) to encrypt our secrets.
 See below for information about how to use `sops`.
 
+We try to keep encrypted secrets files near their configuration files.
+For example, we'll keep a Kubernetes configuration file in the same folder as the secrets that are needed to make changes to that cluster.
+See [the 2i2c GKE cluster folder](https://github.com/2i2c-org/infrastructure/tree/master/config/clusters/2i2c) for an example.
+
 (secrets:sops)=
 ## `sops` overview
 
 [`sops`](https://github.com/mozilla/sops) is a command-line tool for encrypting and decrypting secrets that are on disk.
 It is similar to [`git-crypt`](https://github.com/AGWA/git-crypt) (which is what is used by the Binder SRE team), but gives a bit more visibility into the encrypted fields by only encrypting the *values* rather than the *keys*.
 
-Here's an example of a file that has been encrypted with `sops` (from [our `infrastructure` configuration](https://github.com/2i2c-org/infrastructure/blob/master/config/secrets.yaml)):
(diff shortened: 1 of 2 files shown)


practices/secrets.md
```diff
@@ -10,13 +10,17 @@ This means we try to use services where team members have their own account with
 **When we must share secrets**: If we must share secrets for an account (for example, credentails for deployment to cloud infrastructure), then we use [the command-line tool `sops`](https://github.com/mozilla/sops) to encrypt our secrets.
 See below for information about how to use `sops`.
 
+We try to keep encrypted secrets files near their configuration files.
+For example, we'll keep a Kubernetes configuration file in the same folder as the secrets that are needed to make changes to that cluster.
+See [the 2i2c GKE cluster folder](https://github.com/2i2c-org/infrastructure/tree/master/config/clusters/2i2c) for an example.
```
sgibson91, line 15:
```suggestion
See [the 2i2c GKE cluster folder](https://github.com/2i2c-org/infrastructure/tree/HEAD/config/clusters/2i2c) for an example.
```

This just means that if we decide to rename the default branch of that repo, we don't have to then update all the links in our docs since `HEAD` is a special reference that always points to the latest commit on the default branch

practices/secrets.md
```diff
@@ -10,13 +10,17 @@ This means we try to use services where team members have their own account with
...
+
 (secrets:sops)=
 ## `sops` overview
 
 [`sops`](https://github.com/mozilla/sops) is a command-line tool for encrypting and decrypting secrets that are on disk.
 It is similar to [`git-crypt`](https://github.com/AGWA/git-crypt) (which is what is used by the Binder SRE team), but gives a bit more visibility into the encrypted fields by only encrypting the *values* rather than the *keys*.
 
-Here's an example of a file that has been encrypted with `sops` (from [our `infrastructure` configuration](https://github.com/2i2c-org/infrastructure/blob/master/config/secrets.yaml)):
+Here's an example of a file that has been encrypted with `sops` (from [our `infrastructure` configuration](https://github.com/2i2c-org/infrastructure/blob/master/config/clusters/2i2c/enc-grafana-token.secret.yaml)):
```
sgibson91, line 23:
```suggestion
Here's an example of a file that has been encrypted with `sops` (from [our `infrastructure` configuration](https://github.com/2i2c-org/infrastructure/blob/HEAD/config/clusters/2i2c/enc-grafana-token.secret.yaml)):
```

Same as above

sgibson91 (approved):
LGTM

choldgraf:
Good call @sgibson91 - I need to remember to use `HEAD` more often :-)

Since this doesn't change our strategy, mostly just moves it around, I'll plan to merge this in the next day or two unless there are any suggestions otherwise!

GeorgianaElena (approved):
:rocket:

PR #393, opened 2022-03-16
Review operations and support items first during sprint planning

GeorgianaElena:
This moves the operations and support review before we review any development projects we are working on.
This is to make sure we have enough time to discuss anything that needs discussion and to make sure that we have people assigned to each of them.

Reference: @choldgraf comment [here](https://github.com/2i2c-org/team-compass/issues/390#issuecomment-1068526826).


sgibson91 (approved):
✨

GeorgianaElena:
Thanks @sgibson91 🌼 I will merge this so we can try it for today's meeting and if we don't like it or if it doesn't prove its usefulness, we can revert the PR 🚀

PR #397, opened 2022-03-28
Update time of team meeting in issue template

sgibson91:
I believe we now have these meetings 1 hour later

### DIFF
--- a/.github/ISSUE_TEMPLATE/meeting-team.md
+++ b/.github/ISSUE_TEMPLATE/meeting-team.md
@@ -12,7 +12,7 @@ This is a planning issue for our next team meeting! See below for information ab
 _Information about when and where the meeting will take place_
 
 - **Date:** <YYYY-MM-DD>
-- **Time:** 7:30AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/<YYYY-MM-DD>/07:30/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
+- **Time:** 8:30AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/<YYYY-MM-DD>/08:30/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
 - [**Video conference link**](https://zoom.2i2c.org)
 - [**HackMD for meeting**](https://hackmd.io/Y5SBMxV7R6CMqzeTXgm5kA)
 - [**Calendar for future meetings**](https://calendar.google.com/calendar/embed?src=c_4hjjouojd8psql9i1a8nd1uff4%40group.calendar.google.com&ctz=America%2FLos_Angeles)


.github/ISSUE_TEMPLATE/meeting-team.md
```diff
@@ -12,7 +12,7 @@ This is a planning issue for our next team meeting! See below for information ab
 _Information about when and where the meeting will take place_
 
 - **Date:** <YYYY-MM-DD>
-- **Time:** 7:30AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/<YYYY-MM-DD>/07:30/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
+- **Time:** 8:30AM US/Pacific time (or [**your timezone**](https://arewemeetingyet.com/Los%20Angeles/<YYYY-MM-DD>/08:30/2i2c%20Team%20Meeting#eyJ1cmwiOiJodHRwczovL2hhY2ttZC5pby9ZNVNCTXhWN1I2Q01xemVUWGdtNWtBIn0=))
```
GeorgianaElena, line 15: These `arewemeetingyet` links don't seem to work anymore without a date 😕 
undefined

GeorgianaElena (approved):
I agree pushing this meeting one hour later to better accommodate everybody 🌼 
But, what do people think about discussing whether or not we should still have the meeting be 1:30h long or shoot for keeping it to 60min whenever possible?

.github/ISSUE_TEMPLATE/meeting-team.md
sgibson91: Ping @choldgraf on that - I feel like he got it to work on one point?

choldgraf: argh no I could never figure that out. Didn't we recently discuss **just** using the Google Calendar for this, and having a reminder about it show up in our slacks, rather than using a GitHub issue? I would be fine just using the calendar entry as the source of truth rather than also including the time in our issue template. WDYT?

choldgraf:
Good question on shortening it to 60 minutes. In general, I like the idea of having shorter meetings wherever possible. On the other hand, I worry that we already meet very infrequently and these are rare opportunities to connect with one another and discuss issues that are better discussed synchronously. In the meeting, perhaps the things to discuss are:

1. Is 90 minutes too-long for our monthly team syncs?
2. Is the time we devote to meetings aligned with our needs as a team? (e.g., coordination around work vs. general discussion vs. informal interaction)
3. Is there anywhere we should devote _more_ time in synchronous communication?

Ultimately we want meetings to be useful to our team, so I think our team should be the ones driving these decisions. Would love to hear your thoughts on any of those questions!

.github/ISSUE_TEMPLATE/meeting-team.md
damianavila: Yep, this is totally confusing... having everything on the calendar should help with the SSoT idea...

damianavila:
>  Would love to hear your thoughts on any of those questions!

I do have some thoughts:
1. Let's have a monthly enjoy/informal meeting (30m, NO tech stuff, opt-in, open agenda)
2. Let's have a monthly engineering meeting (1h, GENERAL tech stuff, all hands, ONLY agenda)
And let's those happen in different cycles.

I also have some thoughts related to the sprint process as well but that is another discussion 😉

GeorgianaElena:
I don't think longer meetings compensate for the frequency of them, so I really like Damian's idea!
@damianavila, do you want to pitch your ideas during tommorow's meeting? If so, do you mind adding this to the agenda?

choldgraf:
A few quick things:

- I am a big fan of making time to connect in a way that doesn't involve any agenda (maybe even a promise to not talk about work stuff as much as possible?)
- For our coordination, how about this proposal:
  - We do not use a dedicated issue for these meetings.
  - The 2i2c Team Calendar is the source of truth for this meeting. We can send a reminder to our Slack channel, and other than this we expect team members to keep up-to-date w/ this calendar to know when meetings will happen.
  - It has one link, which is to the meeting agenda/notes.
  - In the agenda/notes, we keep all relevant information about the meeting (dates, links, roles, etc)

damianavila:
I guess this one needs another iteration after yesterday's conversation :wink:

damianavila:
This one is superseded by #402 changes. Closing now.

PR #454, opened 2022-07-04
Convert link checker action to CRON job

choldgraf:
This updates our GitHub action for checking links to be a bit less noisy for each PR. It does these things:

- Each day it runs the action on a CRON job
- If broken links are found via the lychee action, then:
- It re-opens a dedicated link-checking issue with a list of the broken links.

So we can close that issue with PRs to fix the docs, and the issue will automatically re-open in the future if links break again.

Note I will merge this one in so that we can see how it works, might need to iterate a little bit.

### Future improvements

This removes link checking on individual PRs, with the goal of making them less prone to test failures because a non-related link broke. In the future, we could also add functionality to test links only for the files that were modified by the PR, but let's see how this goes first.

### DIFF
--- a/.github/workflows/doc-links.yml
+++ b/.github/workflows/doc-links.yml
@@ -1,13 +1,21 @@
-name: Test documentation for broken links
+name: Build documentation and check links
+
 on:
-  pull_request:
   workflow_dispatch:
+  schedule:
+  # Runs at 1am each day.
+  - cron: '0 1 * * *'
+
+env:
+  BROKEN_LINKS_ISSUE_NUMBER: 453
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
   linkcheck:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-links.yml
```diff
@@ -1,21 +1,71 @@
...
+
 on:
-  pull_request:
   workflow_dispatch:
+  schedule:
+  # Runs at 1am each day.
+  - cron: '0 1 * * *'
+
+env:
+  BROKEN_LINKS_ISSUE_NUMBER: 453
```
choldgraf, line 10:
```suggestion
  # Issue: https://github.com/2i2c-org/team-compass/issues/453
  BROKEN_LINKS_ISSUE_NUMBER: 453
```

damianavila:
> So we can close that issue with PRs to fix the docs, and the issue will automatically re-open in the future if links break again.

From a PM perspective, I think it would be better to have a new issue instead of reopening an old one because most likely that one would need to be unarchived and we would lose quick information about when it was previously closed (and potentially other useful past metadata).

Issue #502, opened 2022-09-01
James on as support steward


choldgraf:
### Context

I spoke with @jmunroe today about ways that we can boost our capacity on community support and our "support stewards" roles. We decided that it would be useful to slot James in now, rather than waiting for him to hit his turn in the rotation, so that he has an opportunity to start learning what the support workflow is like and can help out.

Longer term, we discussed what part of 2i2c should have ownership over the "support strategy" and the system around it. Currently it doesn't really have a home (though is sort-of implicitly in Engineering) - part of our goal in having James serve on this team is to understand if / how the @2i2c-org/partnerships-and-community-guidance team could steward this system instead. 

### Proposal

Here's what I'd propose:

- @choldgraf should take a pass at our current support tickets and help close them out
- Remove me (@choldgraf) from the support stewards team and have @jmunroe join the team
- Have him serve on this team for a few "support steward" cycles, so that he gains experience serving on this team
- After 4-5 weeks, re-discuss the experience and decide on next steps

### Problems we will need to figure out

**We will need to create a way for the support team to assign work to engineers**. If non-engineers like myself, Jim, or James serve on this team more often it is going to create tension when we cannot resolve a technical issue that somebody is having. In those moments it will be crucial that the support team gets _fast_ assistance from engineers to help resolve the issue.

sgibson91:
FYI, it is on my radar to build some scaffolding to make slotting someone into the rotation easier https://github.com/2i2c-org/team-roles-geekbot-sweep/issues/63 I will probably work on this tomorrow as a nice "week closer" task

sgibson91:
If I were to write up onboarding/off-boarding issue templates regarding the Slack user groups and updating the calendar, where would they best live? It would involve running code that lives in the https://github.com/2i2c-org/team-roles-geekbot-sweep repo, but I feel they would be more visible here?

sgibson91:
@choldgraf you have now been removed from the `support-stewards` usergroup on Slack, and hence the pool of future support stewards. Enjoy your last shift. I am incredibly jealous.

@jmunroe has now been added into the rota. Your first shift begins Oct 26th. You will spend two weeks working with Georgiana, and then two weeks working with me (apologies in advance)

sgibson91:
I also opened https://github.com/2i2c-org/team-compass/pull/504 regarding issue templates discussed here: https://github.com/2i2c-org/team-compass/issues/502#issuecomment-1234691433

choldgraf:
[...]

PR #504, opened 2022-09-02
Add issue templates to add/remove a team member from team roles rotation

sgibson91:
ref: https://github.com/2i2c-org/team-compass/issues/502

Add some checklists of steps to take when a team member is taking up or stepping away from a Team Role


GeorgianaElena (approved):
Thanks for adding these. LGTM 🚀

PR #507, opened 2022-09-07
Add finance page

choldgraf:
This adds a finance page to our team compass, as well as a few links to major places where we keep finance information and documentation.

It also adds a link to a little "budget projection" Google Sheet which we can use to project out our finances in the future to help us make decisions.


choldgraf:
Gonna merge since this is just providing links to some pre-existing documents

PR #508, opened 2022-09-08
Replace GitHub with PagerDuty in our Incident Response process

yuvipanda:
PagerDuty is specifically tailored towards handling incidents, so let's use that rather
than try to rig an incident response process on top of GitHub ourselves. I also want us
to focus on the *Incident response team* primarily accessing PagerDuty via their slack
integration, rather than having to swap out to GitHub.

This PR only seeks to replace GitHub with PagerDuty / Slack, and makes no other
changes for now. 

A summary of the changes are:

1. New incidents are *triggered* by creating an Incident in PagerDuty
2. During an incident, the incident response team interacts with the incident
    primarily via slack, although they can do so via the web interface too
3. After the incident, we use the [pagerduty postmortems](https://support.pagerduty.com/docs/postmortems)
    feature to collect information and make a postmortem. This has several features that
    make this process much easier than editing comments on GitHub.

Ref https://github.com/2i2c-org/infrastructure/issues/1118

### DIFF
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -90,7 +90,7 @@ Here is the process that we follow for incidents:
 
    Check the box for "Create a dedicated Public Slack channel for this incident" to create a *new slack channel*
    for discussing the incident. This helps keep chatter off other channels *and* provides an easy location to gather
-   information for the incident report afte the fact.
+   information for the incident report after the fact.
 
    This officially marks the beginning of an incident, and will help make sure we don't accidentally miss steps during
    or after the incident.
@@ -153,7 +153,7 @@ This lets us use notes, status updates from pagerduty as well as messages from S
       processes in the future more easily.
 
       Examples here would be things like `Looked at hub logs with "kubectl logs -n temple -l component=hub" and found <this>` or
-      `Opened the cloud console and discovered notifications about quota". Pasting in commands is very helpful! This is an
+      `Opened the cloud console and discovered notifications about quota`. Pasting in commands is very helpful! This is an
       important way for team members to learn from each other - what you take for granted is perhaps news to someone else, or
       you might learn alternate ways of doing things!
    5. Actions taken to attempt to fix the issue, and their outcome. Paste commands executed if possible, as well as any
@@ -166,7 +166,7 @@ This lets us use notes, status updates from pagerduty as well as messages from S
 6. Fill out the "Analysis" section to the extent possible. In particular, the "Action Items" should be a list with items
    linked out to GitHub issues created for follow-up. Perfection is the enemy of the good here. Save as you go.
 
-7. Click "Save & View Report* when you are done, and ask other members of the incident response team to review the incident report.
+7. Click "Save & View Report" when you are done, and ask other members of the incident response team to review the incident report.
    They might add missing context, additional action items / summary details, or redact information. The person listed as
    the "Owner of the Review Process" is still responsible for making sure the rest of the process is completed.
 
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -56,9 +56,12 @@ Subject Matter Experts
 (incidents:communications)=
 ### Internal communication
 
+- A channel *dedicated* to each incident will be created by pagerduty once an incident is created. This is where most of the
+  discussion about the incident should happen.
 - [`2i2c-org.pagerduty.com`](https://2i2c-org.pagerduty.com/) is a dashboard for managing incidents.
   This is the "source of truth" for any active or historical incidents.
-- [The `#pagerduty-notifications` Slack channel](https://2i2c.slack.com/archives/C041E05LVHB) is where we control PagerDuty and have discussion about an incident. This allows us to have an easily-accessible communication channel for incidents. In general, most interactions with PagerDuty should be via this channel.
+- [The `#pagerduty-notifications` Slack channel](https://2i2c.slack.com/archives/C041E05LVHB) is *primarily* used to trigger
+  new incidents and control pagerduty in other ways. Discussion of *specific* incidents should not happen here.
 
 (support:incident-response)=
 ## Incident response process
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -167,9 +167,10 @@ This lets us use notes, status updates from pagerduty as well as messages from S
 (incidents:postmortem-timeline)=
 ### Writing an incident timeline
 
-Below are some tips and crucial information that is needed for a useful and thorough incident timeline.
+Below are some tips and crucial information that is needed for a useful and thorough incident timeline. You can see
+examples of previous incident reports at the [2i2c-org/incident-reports](https://github.com/2i2c-org/incident-reports/tree/main/reports)
+repository.
 
-% TODO: Add example incident reports for reference when they exist
 The timeline should include:
 
 1. The beginning of the impact.
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -90,7 +90,19 @@ Here is the process that we follow for incidents:
 
 3. **Try resolving the issue** and communicate on the incident-specific channel while you gather information and perform actions - even if only to mark these as notes to yourself.
 4. **Delegate to Subject Matter Experts as-needed**. The Incident Commander is empowered to delegate actions to Subject Matter Experts in order to investigate and resolve the incident quickly.[^note-on-delegation]
-5. **Communicate our status every few hours**. The {term}`External Liason` is expected to communicate incident status and plan with the {term}`Community Representative`s. They should provide periodic updates that describe the current state of the incident, what we have tried, and our intended next steps. Here is a canned response to get started:
+5. **Communicate our status every few hours**. The {term}`External Liason` is
+   expected to communicate incident status and plan with the {term}`Community
+   Representative`s. If the incident commander wants to delegate External Liason duties
+   to someone else, they should:
+
+   1. Assign the *Freshdesk* ticket to the external liason, as that is the default point of
+      communication with community representatives.
+   2. Make a note on the PagerDuty incident as well.
+
+
+   The externl liason should provide periodic updates that describe the current
+   state of the incident, what we have tried, and our intended next steps. Here is
+   a canned response to get started:
 
    ```{button-link} https://2i2c.freshdesk.com/a/admin/canned_responses/folders/80000143608/responses/80000247492/edit
    :color: primary
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -184,6 +184,8 @@ This lets us use notes, status updates from pagerduty as well as messages from S
     the `reports/` directory. This make sure our incidents are all *public*, so
     others can learn from them as well. Given review is already completed in the pagerduty interface, you don't need to wait
     for review to add the report here.
+11. Email a link to the incident report to the community representative, ideally via the Freshdesk ticket used to communicate with
+    them during the incident itself.
 
 
 
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -164,11 +164,13 @@ This lets us use notes, status updates from pagerduty as well as messages from S
       similar incident.
 
 6. Fill out the "Analysis" section to the extent possible. In particular, the "Action Items" should be a list with items
-   linked out to GitHub issues created for follow-up. Perfection is the enemy of the good here.
+   linked out to GitHub issues created for follow-up. Perfection is the enemy of the good here. Save as you go.
 
-7. Review the report, and if the Incident Commander is happy with its completeness, mark the Status dropdown up top as "Reviewed".
+7. Click "Save & View Report* when you are done, and ask other members of the incident response team to review the incident report.
+   They might add missing context, additional action items / summary details, or redact information. The person listed as
+   the "Owner of the Review Process" is still responsible for making sure the rest of the process is completed.
 
-8. Click "Save & View Report" button.
+8. After sufficient review, and if the Incident Commander is happy with its completeness, mark the Status dropdown up top as "Reviewed".
 
 9. Download the PDF, and add it to the `2i2c/infrastrtucture` repository under the `incidents/` directory. This make sure our
    incidents are all *public*, so others can learn from them as well.
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -130,7 +130,7 @@ This lets us use notes, status updates from pagerduty as well as messages from S
 5. **Add Data Sources** that we will use to keep track of the actions that happened around the incident.
    - Link to the slack channel we created for this incident as a "Data Source", filled in with an appropriate time to cover all the messages there.
    - Add any other channels where there was conversation there about the incident (e.g., GitHub Issues or Pull Requests).
-   
+
    Click `Save Data Sources` to populate the timeline below with messages from the slack channels.
 6. **Fill out the timeline**. The goal is to be concise but make it possible for someone reading it to answer "what happened, and when?".
    See [](incidents:postmortem-timeline) for more information.
@@ -140,10 +140,11 @@ This lets us use notes, status updates from pagerduty as well as messages from S
 8. **Click "Save & View Report"** when you are done, and ask other members of the incident response team to review the incident report.
    They might add missing context, additional action items / summary details, or redact information. The person listed as
    the "Owner of the Review Process" is still responsible for making sure the rest of the process is completed.
-9.  After sufficient review, and if the Incident Commander is happy with its completeness, **mark the Status dropdown as "Reviewed"**.
-10. Download the PDF, and add it to the `2i2c/infrastrtucture` repository under the `incidents/` directory. This make sure our incidents are all *public*, so others can learn from them as well.
+9. After sufficient review, and if the Incident Commander is happy with its completeness, **mark the Status dropdown as "Reviewed"**.
+10. Download the PDF, and add it to the [`2i2c/infrastrtucture`](https://github.com/2i2c-org/incident-reports) repository under
+    the `reports/` directory. This make sure our incidents are all *public*, so
+    others can learn from them as well.
 
-% Is there a way to share incidents in a way that doesn't require adding a binary blob to our repository? I think this generates extra toil in a process that already has a lot of toil, and also adds some clunkiness to git-based workflows. For example, could we have a public Google Drive folder where we drag/drop incident reports?
 
 
 (incidents:postmortem-timeline)=
@@ -161,10 +162,10 @@ The timeline should include:
    processes in the future more easily.
 
    For example:
-   
+
    - `Looked at hub logs with "kubectl logs -n temple -l component=hub" and found <this>`
    - `Opened the cloud console and discovered notifications about quota`.
-    
+
    Pasting in commands is very helpful!
    This is an important way for team members to learn from each other - what you take for granted is perhaps news to someone else, or you might learn alternate ways of doing things!
 5. Actions taken to attempt to fix the issue, and their outcome.
@@ -194,7 +195,7 @@ Incident Reports
   A document that describes what went wrong during an incident and what we'll do to avoid it in the future. When we have an {term}`Incident`, we create an Incident Report issue.
 
   This helps us understand what went wrong, and how we can improve our systems to prevent a recurrance. Its goal is to identify improvements to process, technology, and team dynamics that can avoid incidents like this in the future. It is **not** meant to point fingers at anybody and care should be taken to avoid making it seem like any one person is at fault.
-  
+
   This is a *very important* part of making our infrastructure and human processes more stable and stress-free over time, so we should do this after each incident.[^post-mortems].
 ```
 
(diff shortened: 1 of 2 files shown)


yuvipanda:
Me and @jmunroe spent a bunch of time talking about this, and in particular role-playing how this *should* have played out with today's UToronto outage.

A proposed workflow is that after the incident is created, we want pagerduty to evaluate the current local timezone and stated preferences of all engineers and then send everyone who wants to be notified at that point a notification, via methods of their own choosing (SMS, App, Phone call). Engineers can then *acknowledge* the alert if they are able to provide assistance, *or do nothing* if they can't. After an hour (or a configurable time period later!), if nobody has acknowledged the alert, it'll automatically escalate to a (opt-in) group of second tier folks who can respond.

We're going to play with pagerduty rules to try make this possible.

However, even without *any* of this, it still provides value by streamlining the 'incident response' process and removing a medium (GitHub) that currently is a bit of a bottleneck - and instead moving it all to slack.

yuvipanda:
This made me realize that I could reduce the scope of this PR to *just* replacing GitHub with Pagerduty - that alone would've caught that the incident from last week wasn't actually 'over', and would have prevented today's UToronto outage!

yuvipanda:
https://github.com/2i2c-org/infrastructure/issues/1118#issuecomment-1240284103 has follow-up tasks

yuvipanda:
I've tried to follow the incident response suggestions here, and created one for the UToronto outage yesterday https://2i2c-org.pagerduty.com/postmortems/171317d6-5f19-7511-7d3a-117b13f62584

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +75,106 @@ Here is the process that we follow for incidents:
...
+   Assign it to whoever is the **Incident Commander**. This is by default one of the support stewards or whoever is
+   triggering the event, but not necessarily[^note-on-delegation]!
+
+   Provide a descriptive but short Title, but don't sweat it too much!
+
+   If there is a freshdesk ticket for this, provide a link to that in the description.
+
+   Check the box for "Create a dedicated Public Slack channel for this incident" to create a *new slack channel*
+   for discussing the incident. This helps keep chatter off other channels *and* provides an easy location to gather
+   information for the incident report afte the fact.
```
sgibson91, line 93:
```suggestion
   information for the incident report after the fact.
```

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +75,106 @@ Here is the process that we follow for incidents:
...
+
+   1. The beginning of the impact.
+   2. When the incident was brought to our attention, with a link to the source (Freshdesk ticket, slack message, etc).
+   3. When we responded to the incident. This would coincide with the creation of the PagerDuty incident.
+   4. Various debugging actions performed to ascertain the cause of the issue. Talking to yourself as you do this on the
+      slack channel helps a lot here, as it helps communicate your methods to others on the team as well as help improve
+      processes in the future more easily.
+
+      Examples here would be things like `Looked at hub logs with "kubectl logs -n temple -l component=hub" and found <this>` or
+      `Opened the cloud console and discovered notifications about quota". Pasting in commands is very helpful! This is an
```
sgibson91, line 156:
```suggestion
      `Opened the cloud console and discovered notifications about quota`. Pasting in commands is very helpful! This is an
```

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +75,106 @@ Here is the process that we follow for incidents:
...
+      GitHub PRs made. Putting this in Slack again helps.
+   6. Any extra communication from the community affected that helped.
+   7. Whenever the impact was fixed, and how that was verified.
+   8. Whatever else you think would be helpful to someone who finds this incident report a few months from now, trying to fix a
+      similar incident.
+
+6. Fill out the "Analysis" section to the extent possible. In particular, the "Action Items" should be a list with items
+   linked out to GitHub issues created for follow-up. Perfection is the enemy of the good here. Save as you go.
+
+7. Click "Save & View Report* when you are done, and ask other members of the incident response team to review the incident report.
```
sgibson91, line 169:
```suggestion
7. Click "Save & View Report" when you are done, and ask other members of the incident response team to review the incident report.
```

sgibson91 (approved):

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
-  - Confirm that the Incident Commander has the bandwidth and ability to serve in this role.
-  - If not, delegate this to another team member.[^note-on-delegation]
-6. **Designate an {term}`External Liason`**. Do this in the Incident issue. By default, this is the Incident Commander, though they may delegate this to others.[^note-on-delegation]
-7. **Investigate and resolve the incident**. The Incident Commander should follow the structure of the incident issue opened in the step above.
-8. **Delegate to Subject Matter Experts as-needed**. The Incident Commander is empowered to delegate actions to Subject Matter Experts in order to investigate and resolve the incident quickly.[^note-on-delegation]
-9. **Communicate our status every few hours**. The {term}`External Liason` is expected to communicate incident status and plan with the {term}`Community Representative`s. They should provide periodic updates that describe the current state of the incident, what we have tried, and our intended next steps. Here is a canned response to get started:
+2. **Trigger an incident in PagerDuty**. Below are instructions for doing so via [the 2i2c slack](incidents:communications).
```
choldgraf, line 79: We should have a team learning session for how to use the PagerDuty interface. Could somebody record themselves going through the full PagerDuty workflow as part of these instructions? (maybe we could use something like https://loom.com/ ?)

projects/managed-hubs/incidents.md
```diff
@@ -53,11 +53,12 @@ Subject Matter Experts
...
 
+(incidents:communications)=
 ### Internal communication
 
-- The Slack channel [{guilabel}`#support-freshdesk`](https://2i2c.slack.com/archives/C028WU9PFBN) contains real-time communication about support issues. Use this to signal-boost support requests related to {term}`Incidents`.
-- [Issues with the {guilabel}`incident` label](https://github.com/2i2c-org/infrastructure/issues?q=is%3Aopen+label%3A%22type%3A+Hub+Incident%22+sort%3Aupdated-desc) are where we track progress when [resolving incidents](support:incident-response).
-
+- [`2i2c-org.pagerduty.com`](https://2i2c-org.pagerduty.com/) is a dashboard for managing incidents.
+  This is the "source of truth" for any active or historical incidents.
+- [The `#pagerduty-notifications` Slack channel](https://2i2c.slack.com/archives/C041E05LVHB) is where we control PagerDuty and have discussion about an incident. This allows us to have an easily-accessible communication channel for incidents. In general, most interactions with PagerDuty should be via this channel.
```
choldgraf, line 61: Ah we should also add a bullet-point for `incident-specific` channels since we create those below.

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
+   - **Add a link to the FreshDesk ticket** in the description (if there is one).
+   - **Create a new Slack channel** by checking the box for `Create a dedicated Public Slack channel for this incident`.
+     Use this channel for all conversations about the incident.
+
+   This officially marks the beginning of an incident, and will help make sure we don't accidentally miss steps during or after the incident.
+
+3. **Try resolving the issue** and communicate on the incident-specific channel while you gather information and perform actions - even if only to mark these as notes to yourself.
+4. **Delegate to Subject Matter Experts as-needed**. The Incident Commander is empowered to delegate actions to Subject Matter Experts in order to investigate and resolve the incident quickly.[^note-on-delegation]
+5. **Communicate our status every few hours**. The {term}`External Liason` is expected to communicate incident status and plan with the {term}`Community Representative`s. They should provide periodic updates that describe the current state of the incident, what we have tried, and our intended next steps. Here is a canned response to get started:
```
choldgraf, line 93: How do we define the External Liason? Before we said "use the issue" for this but I don't think there's a way to explicitly say this in PagerDuty

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
-9. **Communicate when the incident is resolved**. When we believe the incident is resolved, communicate with the Community Representative that things should be back to normal. Mark the FreshDesk ticket as {guilabel}`Resolved`.
-10. **Fill in the {term}`Incident Report`**. The Incident Commander should do this in partnership with the Incident Response Team.
-11. **Close the incident ticket**. Once we have confirmation from the community (or no response after 48 working hours), and have filled in the incident {term}`Incident Report`, then close the incident by:
-    - Closing the incident issue on GitHub
-    - Marking the FreshDesk ticket as {guilabel}`Closed`
+6. **Communicate when the incident is resolved**. When we believe the incident is resolved, communicate with the Community Representative that things should be back to normal.
+   - Mark the incident as "Resolved" in pagerduty.
+   - Mark the FreshDesk ticket as {guilabel}`Closed`.
+7. **Create an incident report**.
+   See [](incidents:create-report) for more information.
```
choldgraf, line 105: I moved all of this into a dedicated section so it didn't clutter these to-do lists too much.

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
 
 [^note-on-delegation]: If you cannot find somebody to take on this work, or feel uncomfortable delegating, the {term}`Project Manager` should help you, and is empowered to delegate on your behalf.
 
+(incidents:create-report)=
+## Create an Incident Report
+
+Once the incident is resolved, we must create an {term}`Incident Report`.
+The **Incident Commander** is responsible for making sure the Incident Report is completed, even though they may not be the person doing it.
```
choldgraf, line 113: When can the incident commander assign reporting duties to somebody else? I think if we don't define this then they will always be the ones that end up doing it. If we have this documented elsewhere we should cross-link it

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
+
+We practice a [blameless culture](https://www.blameless.com/sre/what-are-blameless-postmortems-do-they-work-how) around incident reports.
+Incidents are **always** caused by systemic issues, and hence solutions must be systemic too.
+Go out of your way to make sure there is no finger-pointing.
+
+We use [PagerDuty's postmortem feature](https://support.pagerduty.com/docs/postmortems) to create the Incident Report.
```
choldgraf, line 119: This documentation suggests that this requires a paid plan. Is that true for us? If so we should make an explicit decision that we wish to pay for this: https://support.pagerduty.com/docs/postmortems

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
+   Perfection is the enemy of the good here. Save as you go.
+8. **Click "Save & View Report"** when you are done, and ask other members of the incident response team to review the incident report.
+   They might add missing context, additional action items / summary details, or redact information. The person listed as
+   the "Owner of the Review Process" is still responsible for making sure the rest of the process is completed.
+9.  After sufficient review, and if the Incident Commander is happy with its completeness, **mark the Status dropdown as "Reviewed"**.
+10. Download the PDF, and add it to the `2i2c/infrastrtucture` repository under the `incidents/` directory. This make sure our incidents are all *public*, so others can learn from them as well.
+
+% Is there a way to share incidents in a way that doesn't require adding a binary blob to our repository? I think this generates extra toil in a process that already has a lot of toil, and also adds some clunkiness to git-based workflows. For example, could we have a public Google Drive folder where we drag/drop incident reports?
```
choldgraf, line 146: Signal-boosting this question from when I was making edits. What do you think about a Google Drive folder that exists just for the purpose of sharing incident reports?

If not that, maybe we could have a dedicated documentation site for this instead of using `infrastructure/`?

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +76,104 @@ Here is the process that we follow for incidents:
...
+
+
+(incidents:postmortem-timeline)=
+### Writing an incident timeline
+
+Below are some tips and crucial information that is needed for a useful and thorough incident timeline.
+
+The timeline should include:
```
choldgraf, line 154: I think the easiest thing we can do is link to pre-existing timelines that are well-written. Then people can just riff off of the structure of those

choldgraf (commented):
This seems like a really helpful service to use in order to provide some structure to our incident response process. In general, it looks good to me. I had a few comments and suggestions throughout.

One concern I have is that this is a fairly involved process. How can we ensure that we reliably and diligently follow this process?

Argh I took a pass and added some suggested edits to up the lists into different sections, but accidentally pushed to your branch instead of making a PR. Happy to discuss and I'll provide some comments below to focus discussion.

projects/managed-hubs/incidents.md
yuvipanda: I'm ok with it being in a different repo. Let's just do that, I don't want to use Google Drive for this - it's far less public.

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf i agree! once we have a few more incidents we can incorporate that here.

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf I have:

1. created https://github.com/2i2c-org/incident-reports
1. created https://github.com/2i2c-org/infrastructure/pull/1703 to move the existing 1 incident report we have in infrastructure repo to there
2. modified this PR to point to that.

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf I opened https://github.com/2i2c-org/meta/issues/374

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf tried to clarify this.

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf I've updated with some suggestions

projects/managed-hubs/incidents.md
yuvipanda: Done, and reworded this a little

yuvipanda:
Thanks a lot @sgibson91 and @choldgraf!

@choldgraf I think I've addressed all your concerns.

Re: learning to use pagerduty, there is a lot of documentation they maintain at https://university.pagerduty.com/ - would that be enough? I'm also happy to try make a video, but do not want to block this PR on that.

This process IMO is actually *simpler* than what we have now, because:

1.  *during* an incident, all interactions happen via slack (which we already use!)
2. Constructing a timeline becomes *much* easier - it's a fully manual process now
3. Easier to see when an incident can be marked as 'completed', a bit more difficult with GitHub now.

We will adjust this as we go along! However, I wrote this document so obviously I'll feel this way :D Would love to hear from others if they think this is too complex!

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +79,119 @@ Here is the process that we follow for incidents:
...
+   See [](incidents:create-report) for more information.
 
 [^note-on-delegation]: If you cannot find somebody to take on this work, or feel uncomfortable delegating, the {term}`Project Manager` should help you, and is empowered to delegate on your behalf.
 
+(incidents:create-report)=
+## Create an Incident Report
+
+Once the incident is resolved, we must create an {term}`Incident Report`.
+The **Incident Commander** is responsible for making sure the Incident Report is completed, even though they may not be the person doing it.
+If they are *not* the person doing it, they should still creat the incident report, but assign `Owner of the Review Process`
```
GeorgianaElena, line 129:
```suggestion
If they are *not* the person doing it, they should still create the incident report, but assign `Owner of the Review Process`
```

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +79,119 @@ Here is the process that we follow for incidents:
...
+   Talking to yourself as you do this on the slack channel helps a lot here, as it helps communicate your methods to others on the team as well as help improve
+   processes in the future more easily.
+
+   For example:
+
+   - `Looked at hub logs with "kubectl logs -n temple -l component=hub" and found <this>`
+   - `Opened the cloud console and discovered notifications about quota`.
+
+   Pasting in commands is very helpful!
+   This is an important way for team members to learn from each other - what you take for granted is perhaps news to someone else, or you might learn alternate ways of doing things!
```
GeorgianaElena, line 187: ❤️

GeorgianaElena (approved):
> We should have a team learning session for how to use the PagerDuty interface. Could somebody record themselves going through the full PagerDuty workflow as part of these instructions? (maybe we could use something like https://loom.com/ ?)

Maybe a demo during the next monthly team meeting?

projects/managed-hubs/incidents.md
choldgraf: Nice! One question: what if we began by using PagerDuty for the incident itself, and GitHub to define the post-mortem content?

The reason I mention this is because (aside from the cost questions), it might solve two other problems:

1. Making the post-mortem easily accessible. If we used GitHub issues for this (as we do now), it is quite easy to search through all of our historical records of post-mortems.
2. Tying the post-mortem to the issues we create. The post-mortem almost always involves the creation of a bunch of new github issues etc for follow-up, and if we used GitHub for the post-mortem content itself, it would allow us to cross-reference to issues / PRs more easily.

If over time it seems like this doesn't work for us, we could always upgrade to a business plan.

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +79,119 @@ Here is the process that we follow for incidents:
...
+    the `reports/` directory. This make sure our incidents are all *public*, so
+    others can learn from them as well.
+
+
+
+(incidents:postmortem-timeline)=
+### Writing an incident timeline
+
+Below are some tips and crucial information that is needed for a useful and thorough incident timeline.
+
```
choldgraf, line 171:
```suggestion

% TODO: Add example incident reports for reference when they exist
```

projects/managed-hubs/incidents.md
```diff
@@ -75,34 +79,119 @@ Here is the process that we follow for incidents:
...
+5. **Communicate our status every few hours**. The {term}`External Liason` is
+   expected to communicate incident status and plan with the {term}`Community
+   Representative`s. If the incident commander wants to delegate External Liason duties
+   to someone else, they should:
+
+   1. Assign the *Freshdesk* ticket to the external liason, as that is the default point of
+      communication with community representatives.
+   2. Make a note on the PagerDuty incident as well.
+
+
```
choldgraf, line 105:
```suggestion
```

projects/managed-hubs/incidents.md
choldgraf: This looks good to me, but I think an unanswered question is **when and why** would a person other than the incident commander be the one to fill this out? Should we encourage the engineer most-involved in resolving the incident to fix this? Should we try to rotate between members? Should we encourage the incident commander to do this most of the time but only ask somebody else if they really can't do it in a timely fashion?

projects/managed-hubs/incidents.md
sgibson91: > If we used GitHub issues for this (as we do now), it is quite easy to search through all of our historical records of post-mortems.

I would push back on this being "easy" a little bit. We do have to sift them out from all of the other issues going on in the infrastructure repo, and it's a very busy repo. We could transfer those issues to https://github.com/2i2c-org/incident-reports though.

choldgraf: When I do this I just filter by the Hub Incident label. At least this restricts to only the "incident" issues...agree it would be nicer if it were like a little Jupyter Book site or something.

- https://github.com/2i2c-org/infrastructure/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22type%3A+Hub+Incident%22

yuvipanda:
@choldgraf IMO, the postmortem process is one of the *main* reasons to use pagerduty, and I'd really like to try it out this way! In the future, I think automated alerts (and alerting / escalation) should also go through here I think

projects/managed-hubs/incidents.md
yuvipanda: @choldgraf I think our current experiences have been that all the 3 roles have been basically the same person, and I think our wide timezone spread has made it difficult for delegation to happen as well. I think this is an important question to answer, but heavily constrained by our timezone setup. Do you think this should block this PR?

projects/managed-hubs/incidents.md
yuvipanda: I agree with @sgibson91! IMO, it gets completely lost in GitHub and it's really difficult to find - especially as it is an issue comment that gets edited. It is also an incredibly manual process right now, and after trying out the process once in pagerduty I think using it is going to make postmortems *far more likely*. Also you can go to pagerduty and see a list of all incident reports - https://2i2c-org.pagerduty.com/postmortems.

In the future, I think we can use pagerduty for escalations and automated alerts as well. So regardless of what we do, I really want to move *away* from 'edit a github issue comment' as the model for how we do incident reports.

choldgraf:
Sounds good - I defer to the @2i2c-org/tech-team's wishes on this one. If y'all think this is the right system then we should go for it.

yuvipanda:
@choldgraf I've added a link to examples too

choldgraf (approved):
I think that we are all in agreement that this is a good direction to move towards. We've also already paid for a years-worth of accounts for PagerDuty. I think that we should merge this PR in and begin iterating from there.

yuvipanda:
I agree @choldgraf! Merged!

PR #509, opened 2022-09-08
Update shared drive link

choldgraf:
This updates our shared drive link.

- ref: https://github.com/2i2c-org/meta/issues/353


choldgraf:
Gonna merge since this is just updating outdated docs links

Issue #513, opened 2022-09-13
Update our organizational mission


choldgraf:
### Context

We've [defined our organizational mission in this Google Doc](https://docs.google.com/document/d/1qlKC2GcZlkTJfu6SWjJJyV6nUKEAJeUbhD7hA8OKcFQ/edit#heading=h.fta9cs8llgn8) and also have some mission/values information listed in the [Our Values and Principles](https://2i2c.org/about/) section of the About page (in fact our [team compass points to this page](https://team-compass.2i2c.org/en/latest/about/index.html)).

Here is the language we use there:

> Our mission is to make research and education more impactful, accessible, and delightful by developing, operating, and supporting infrastructure for interactive computing.

It was defined in the early days of 2i2c, and we now have a larger team than before, and our thinking about our mission has sharpened over time. As we are taking stock of our mission, strategy, and organizational structure, let's use this as an opportunity to refine our mission[^1].

[^1]: There are many ways to define a mission statement, [here's one useful article](https://www.process.st/company-mission-statement/). There's no one right way to do it, but I like the following criteria: (1) short, (2) simple, (3) evokes specific actions or intentions, (4) Makes clear what you do and don't do.

### Proposal

We should reflect on the mission that we have listed, with the goal of proposing any modifications that we'd like to make. Ultimately, we should have a mission statement that has the following qualities.

- Is explicit, and defined in our Team Compass rather than in 2i2c.org/about
- Is something we all agree is a mission we can get behind

colliand:
The discussion in Paris and captured in https://github.com/2i2c-org/meta/issues/583 highlighted orthogonal dimensions in 2i2c's mission: **breadth** and **depth**. I found this useful and suggest we find ways to incorporate that language into an improvement of 2i2c's current mission statement:

>To make research and education more impactful, accessible, and delightful by developing, operating, and supporting infrastructure for interactive computing.

The breadth X depth phrasing from @choldgraf in the retreat summary and our discussions about build-measure-learn loops led me to think of this objective function:
$$G=\sum_{Communities}Breadth \cdot Depth$$

2i2c advances our mission objective by growing the goodness function $G$. 2i2c's goodness grows by
1. expanding the number of communities we serve
2. expanding the breadth within a community
3. expanding the depth within a community

2i2c enables more (breadth) people to participate in data+compute-intensive (depth) science/scholarship/policy.

[...]

PR #516, opened 2022-09-14
Reorganize organization section documentation

choldgraf:
This is a slight re-organization of our "organizational" sections in the Team Compass. The goal is to highlight these aspects of 2i2c at the top since it is the policy that guides everything else in the organization. It also should make certain information like our mission and values easier to find and more clearly defined. Here are the major updates:

- Rename `about/` to `organization/` since these are org-wide documents.
- Move it to the top of our TOC so it's the first thing in our team compass.
- Breaking up pages a bit more to have dedicated pages for things rather than nested sub-sections on a page.
- A few updates and cleaning up content that has become out of date or was superfluous

A few issues that inspired this, I'll keep churning on these once we get this merged:

- ref: https://github.com/2i2c-org/team-compass/issues/513
- ref: https://github.com/2i2c-org/meta/issues/166


sgibson91 (approved):

choldgraf:
This doesn't change a ton of content, mostly re-organizes things, so I'll merge and we can focus future discussion on the content-focused stuff.

Issue #532, opened 2022-10-10
Move our New Hub process documentation to `infrastructure/` and update it to reflect a two-template process


choldgraf:
### Context

In a recent PR[^1] we split our `New Hub` template into two different templates

- [the request information template](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=type%3A+hub&template=3_new_hub-request-info.yaml&title=%5BRequest+Information%5D+New+Hub%3A+%7B%7B+HUB+NAME+%7D%7D)
- [the collect information template](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=type%3A+hub&template=2_new-hub-collect-info.yml&title=%5BCollect+Information%5D+New+Hub%3A+%7B%7B+HUB+NAME+%7D%7D)

However, we have not updated the process in our Source of Truth, and there is now a mismatch. [Here is where we document our new hub process](https://team-compass.2i2c.org/en/latest/projects/managed-hubs/sales.html#steps-in-selling-new-services).

[^1]: - https://github.com/2i2c-org/infrastructure/pull/1733

### Proposal

I propose that we do 2 things:

1. **Move the `New Hub` process documentation to our `infrastructure/` repository**. We can cross-link to those docs from here. My reasoning for this is that the process for setting up a new hub should be "owned" by the engineering team. It will be easier for that team to own this process if we delegate "source of truth status" to that location.
2. **Clarify what these two templates mean**. Provide some explanatory information for these two templates and how they work together in a single process. After looking again, I am not sure what is the difference between `collect information` and `request information`, and I think we should disambiguate these two.

choldgraf:
@GeorgianaElena since you updated these templates in `infrastructure/` would you be willing to write the proper docs there to document this new process, and then I can update the `team-compass` docs to link to it?

PR #534, opened 2022-10-12
Update process about opening a new hub issue

GeorgianaElena:
Together with https://github.com/2i2c-org/infrastructure/pull/1761, closes https://github.com/2i2c-org/team-compass/issues/532.

I personally believe that this info should still reside in this repository because the eng team are the ones responsible for pushing forward the `provide` issue, which has enough info to deploy the new hub and the `request` one should be "owned" by the leads team?


projects/managed-hubs/sales.md
```diff
@@ -78,7 +78,9 @@ Set up the service
...
+  Feel free to ping the engineering team whenever their input is needed.
+  - Make sure the step above concluded with a [**new {guilabel}`New Hub - Provide Information` issue**](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=type%3A+hub&template=2_new-hub-provide-info.yml&title=%Provide+Information%5D+New+Hub%3A+%7B%7B+HUB+NAME+%7D%7D) and this issue is appropriately added to the Team Backlog. This will alert the engineering team that it is time to create a new hub for this community! An engineer will then be assigned to a `New Hub - Provide Information` issue in order to deploy this hub and not a `New Hub - Request Information` issue.
```
choldgraf, line 83: A quick question: could we call this issue template `Request a new hub` instead of `Provide information`? The goal of that issue is to create a hub right? And we provide information as part of that process. I think this could make it easier to distinguish from the "Request information" issue.

GeorgianaElena: I believe this is a very good idea!

GeorgianaElena: @choldgraf, I've renamed it and also opened https://github.com/2i2c-org/infrastructure/pull/1797

choldgraf (approved):
This looks good to me - merging this in and we can iterate from there!

PR #536, opened 2022-10-13
Add accounting information and links

choldgraf:
This adds a number of links and information about our accounting and finances documents. Its goal is to make it easier for people to quickly find things, and to understand the relevant information to consider.

There are basically two different types of resources this links out to:

- CS&S accounting information, including all of our contracts and our monthly financial statements
- 2i2c accounting information, including our informal contract documents and budget *projections*

### Thinks to consider for review

- Is there any accounting questions you've had in the last few months that isn't answered by one of these links?
- Is it clear how to access the various kinds of documents we might need to know about for accounting purposes?
- Is it OK to have this all on one page or should we split it up into sub-pages at this point?

### DIFF
--- a/finance/index.md
+++ b/finance/index.md
@@ -2,6 +2,11 @@
 
 Main documentation for 2i2c's financial and accounting information.
 
+:::{admonition} A note on access
+All 2i2c-specific accounting documents should be accessible on our Team Drive to anybody with a `@2i2c.org` address.
+We don't control the access conditions for the CS&S documents described below, so if you don't have access to these, ask the Executive Director.
+:::
+
 ## General information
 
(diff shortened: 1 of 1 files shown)


choldgraf:
It is probably most-relevant for @colliand and @jmunroe who are helping think about the sustainability and partnerships side of things.

yuvipanda:
I FUCKING LOVE THIS

choldgraf:
I'm going to self-merge this one because I think it is largely "informational" rather than setting new policy. I'm happy to iterate and improve the explanations here if people run into confusing things!

damianavila:
This is great content, @choldgraf! Thanks.

PR #568, opened 2022-11-16
Refactor job titles to be underneath functional areas

choldgraf:
This makes a few improvements to our jobs / hiring documentation to make it easier to navigate. Here are the main changes:

- Updated links to our hiring template documentation and process
- **Moved** the definitions of our job titles / descriptions **outside of the `People` group** and **into each functional area**. My rationale for this is that the members of each area are the ones that define the roles/responsibilities, etc, and so it makes sense for those roles to exist within the groups rather than in a generic "`People group`" section.
- The `People` group now defines the policy / structure that job descriptions must follow, but not the role text itself


choldgraf (commented):
Just FYI I'll plan on merging this in a day or so unless there are objections. I don't believe this is changing any policy, just describing things more explicitly and organizing

choldgraf:
I'm gonna merge this one in since it is primarily a re-organization and nobody has objected so far. I think this will make the docs easier to navigate and to know where these various roles "belong"

PR #572, opened 2022-11-21
Move strategy docs here from service

choldgraf:
This moves several strategy / planning documents into the Team Compass, as they used to be in docs.2i2c.org, and this made it hard to keep track of them. We can decide to re-delegate/move these elsewhere in the future, but for now I think we should keep them in the TC since it is easier to spot as a source of truth.

### References

- Sibling to https://github.com/2i2c-org/docs/pull/174


sgibson91 (approved):

sgibson91:
This sibling is merged, so let's merge this too

PR #575, opened 2022-11-23
Make reimbursement process more visible

choldgraf:


choldgraf:
gonna self-merge since this is a minor one

PR #580, opened 2022-12-01
Add "end of year" time off policy

choldgraf:
This refines some of our time off policies around the "expected downtime" at the end of the year. It tries to focus this around setting expectations rather than forcing people to take one action or another. 

Some notes from our conversation about this in the team strategy meeting are here: https://docs.google.com/document/d/1HoNX8T8IQ1uhS2ryi1r9iS-nSbPT1b1Y7HsjosbHme8/edit#heading=h.24ru6ob0hmvh

### DIFF
--- a/people/time-off.md
+++ b/people/time-off.md
@@ -57,11 +57,7 @@ Team members can use this time as they wish (including continuing to work), but 
 Here is our policy during this time:
 
 - All team meetings are cancelled during this time.
-- We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
-- If there is an incident, we will take steps to minimize damage and harm, and will plan time _after this period_ to resolve underlying problems.
-- We do not perform any non-essential cloud changes for communities.
-
-See [](support:expected-time-off) for our support changes during this time.
+- [](support:expected-time-off:policy) for our support policy during this time.
 
 ## Unknowns to resolve
 
--- a/people/time-off.md
+++ b/people/time-off.md
@@ -45,7 +45,7 @@ In this case, our guidelines around letting others know ahead of time are somewh
 Just do the best you can, and try to balance your own constraints with respect for the team’s bandwidth and planning.
 
 (time-off:annual-expected)=
-## End-of-year expected time off
+## Expected end-of-year time off
 
 The **final 14 days and the first 3 days of the year are "expected time off" for all of 2i2c**.
 
--- a/people/time-off.md
+++ b/people/time-off.md
@@ -47,7 +47,7 @@ Just do the best you can, and try to balance your own constraints with respect f
 (time-off:annual-expected)=
 ## End-of-year expected time off
 
-The **final 14 days of the year are "expected time off" for all of 2i2c**.
+The **final 14 days and the first 3 days of the year are "expected time off" for all of 2i2c**.
 
 During this time our organization operates at a reduced capacity and with greatly diminished commitments to our {external+docs:ref}`service level objectives <objectives:stability>`.
 
@@ -58,8 +58,8 @@ Here is our policy during this time:
 
 - All team meetings are cancelled during this time.
 - We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
-- We do not perform any non-essential cloud changes for communities.
 - If there is an incident, we will take steps to minimize damage and harm, and will plan time _after this period_ to resolve underlying problems.
+- We do not perform any non-essential cloud changes for communities.
 
 See [](support:expected-time-off) for our support changes during this time.
 
--- a/people/time-off.md
+++ b/people/time-off.md
@@ -44,6 +44,25 @@ Follow the process described above in this case - you are welcome to describe yo
 In this case, our guidelines around letting others know ahead of time are somewhat relaxed, as sick leave may happen unexpectedly.
 Just do the best you can, and try to balance your own constraints with respect for the team’s bandwidth and planning.
 
+(time-off:annual-expected)=
+## End-of-year expected time off
+
+The **final 14 days of the year are "expected time off" for all of 2i2c**.
+
+During this time our organization operates at a reduced capacity and with greatly diminished commitments to our {external+docs:ref}`service level objectives <objectives:stability>`.
+
+Our support and operational roles are still staffed, but with greatly reduced expectations around responsiveness and resolution.
+Team members can use this time as they wish (including continuing to work), but we wish to set the _expectation_ with internal and external partners that we will be less responsive.
+
+Here is our policy during this time:
+
+- All team meetings are cancelled during this time.
+- We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
+- We do not perform any non-essential cloud changes for communities.
+- If there is an incident, we will take steps to minimize damage and harm, and will plan time _after this period_ to resolve underlying problems.
+
+See [](support:expected-time-off) for our support changes during this time.
+
 ## Unknowns to resolve
 
 - This process is likely too informal if our team significantly scales in size and complexity.
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -247,6 +247,8 @@ Below is our policy for support during expected time off:
   % NOTE: In the future we may define a policy for support if a community knows and
   %   tells us they are doing essential work during expected time off.
   %   We can update this policy when we discuss and make a decision on this.
+  %
+  %   Some conversation in: https://github.com/2i2c-org/meta/issues/435
 - We will take steps to minimize harm and avoid catastrophic problems (e.g. incidents that drastically increase costs), but will not perform non-essential actions. We will assume there is _no essential work happening on any of our hubs for all communities_.
 - If there is a catastrophic incident, we will take the minimal number of actions to reduce risk and damage to a reasonable level.
 - For non-catastrophic incidents and general change requests, we will wait until _after this period_ to resolve them.
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -239,12 +239,15 @@ There are some cases when we intentionally reduce our operations and support cap
 See [](time-off:annual-expected) for our broader policies and support commitments during this time.
 
 (support:expected-time-off:policy)=
-### Expected time off policy
+### Expected time off support policy
 
 Below is our policy for support during expected time off:
 
 - We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
-- We will take steps to minimize harm and avoid catastrophic problems (e.g. incidents that drastically increase costs), but will not perform non-essential actions. We will not treat "a general outage" as catastrophic unless we get an explicit signal in advance from a community representative they are doing essential work during this time.
+  % NOTE: In the future we may define a policy for support if a community knows and
+  %   tells us they are doing essential work during expected time off.
+  %   We can update this policy when we discuss and make a decision on this.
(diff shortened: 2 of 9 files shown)


choldgraf:
I invited review from the people that I think I remember being there, but I might be mis-remembering so please let me know if I got that wrong

people/time-off.md
```diff
@@ -44,6 +44,25 @@ Follow the process described above in this case - you are welcome to describe yo
...
+## End-of-year expected time off
+
+The **final 14 days of the year are "expected time off" for all of 2i2c**.
+
+During this time our organization operates at a reduced capacity and with greatly diminished commitments to our {external+docs:ref}`service level objectives <objectives:stability>`.
+
+Our support and operational roles are still staffed, but with greatly reduced expectations around responsiveness and resolution.
+Team members can use this time as they wish (including continuing to work), but we wish to set the _expectation_ with internal and external partners that we will be less responsive.
+
+Here is our policy during this time:
```
damianavila, line 57: I really like this list setting clear expectations for ourselves and external parties.

people/time-off.md
```diff
@@ -44,6 +44,25 @@ Follow the process described above in this case - you are welcome to describe yo
 In this case, our guidelines around letting others know ahead of time are somewhat relaxed, as sick leave may happen unexpectedly.
 Just do the best you can, and try to balance your own constraints with respect for the team’s bandwidth and planning.
 
+(time-off:annual-expected)=
+## End-of-year expected time off
+
+The **final 14 days of the year are "expected time off" for all of 2i2c**.
```
damianavila, line 50: Usually, people do not start until the 3rd of Jan (maybe more?), maybe we should include those as well, so people are not expecting us to start working on Jan 1st?

choldgraf: How about `**All of 2i2c has "expected time off" from December 17th through January 3rd of the following year.**`

people/time-off.md
```diff
@@ -44,6 +44,25 @@ Follow the process described above in this case - you are welcome to describe yo
...
+
+Our support and operational roles are still staffed, but with greatly reduced expectations around responsiveness and resolution.
+Team members can use this time as they wish (including continuing to work), but we wish to set the _expectation_ with internal and external partners that we will be less responsive.
+
+Here is our policy during this time:
+
+- All team meetings are cancelled during this time.
+- We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
+- We do not perform any non-essential cloud changes for communities.
+- If there is an incident, we will take steps to minimize damage and harm, and will plan time _after this period_ to resolve underlying problems.
```
sgibson91, line 62: I personally would rearrange this list so that the two bullet points that discuss incidents are next to each other, since they are related. But I'm being super picky here. Don't block merge on this.

projects/managed-hubs/support.md
```diff
@@ -231,3 +231,35 @@ Support Budget
...
+
+  > **Note:** the 2i2c team is operating at an **expected reduced capacity** until `ENDDATE`.
+  > We will provide operational support for significant cloud incidents and outages, but not for non-essential questions or change requests.
+  > We will be less responsive than normal, and will return to answer your questions and resolve issues after the time off.
+
+% TODO: In the future, we want to add an extra step:
+%
+% Send an e-mail to our mailing list for Community Representatives communicating our intent to be on expected reduced capacity.
+% 
+% See issue: https://github.com/2i2c-org/team-compass/issues/579
```
sgibson91, line 261: Two extra steps! The banner on our hubs as well!

people/time-off.md
sgibson91: I would try to avoid specific dates because this will fluctuate each year, so maybe?:

`final 14 days and first 3 days of the year(s)`

sgibson91 (commented):
This looks good to me, but I think we should get whole team consensus on this one, particularly around the list of expected actions starting on L59. We need to make sure that all team members agree that that is the level of service we are providing during this period, even if (especially if) individual plans are not necessarily "I'm not working this period".

choldgraf:
Thanks for the comments - I've tried to clarify and incorporate each of your suggestions above, thanks!

people/time-off.md
yuvipanda: I like this list

yuvipanda (approved):

people/time-off.md
```diff
@@ -44,6 +44,25 @@ Follow the process described above in this case - you are welcome to describe yo
...
+Team members can use this time as they wish (including continuing to work), but we wish to set the _expectation_ with internal and external partners that we will be less responsive.
+
+Here is our policy during this time:
+
+- All team meetings are cancelled during this time.
+- We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
+- If there is an incident, we will take steps to minimize damage and harm, and will plan time _after this period_ to resolve underlying problems.
+- We do not perform any non-essential cloud changes for communities.
+
+See [](support:expected-time-off) for our support changes during this time.
```
GeorgianaElena, line 64: I understand the following about the responsibility of a support steward during these expected reduced availability periods:

*The support steward will monitor the support email during this time and will react to incidents by doing their best to ensure hub is up and available*

Did I understand the list correctly?

GeorgianaElena (approved):
This generally looks good to me!

*see https://github.com/2i2c-org/meta/issues/435* also

people/time-off.md
sgibson91: No, I think during this period, the support steward should do their best to make sure the hub is "safe", by which I mean there's no cryptominers, etc. We should divert efforts to make sure the hub is up and available and fixing issues for the long-term to when the whole team are back

choldgraf:
### Update: added a FreshDesk footer template

- I've added a FreshDesk footer template in our FreshDesk account to note expected downtime. It should show up in your replies until we disable it!
- I've also added docs to this PR showing how to do it. I've put the `freshdesk.md` documentation under `Community guidance`. I think we'll need to decide in the future if it should instead live under `Engineering`.

people/time-off.md
choldgraf: I think the Engineering team should decide what kind of promises they'd like to make, I think our two choices are basically:

- We'll only provide fixes for catastrophic problems (e.g. crypto mining burning through credits) but won't treat "a general outage" as catastrophic unless we get an explicit signal from a community representative. OR:
- We'll keep hubs up-and-running during expected downtime.

sgibson91: I think we should aim for the first point. Most of our communities are in the global north and will therefore be in a similar reduced capacity too, so I definitely think a hub outage during this period is not catastrophic.

damianavila (approved):
LGTM

choldgraf:
OK I've just pushed a commit that does a couple things:

1. Moves our support policy to the `support.md` document and cross-links it from our `people/` document. That way our support policy is defined closer to our support docs, which I think is an easier way to track it.
2. Makes our support policy during this time more explicit, per conversation above.

It implements the suggestions that y'all approved above, so if I don't hear any negative feedback or suggestions I'll assume we're all still in approval.

projects/managed-hubs/support.md
```diff
@@ -231,3 +231,49 @@ Support Budget
...
+There are some cases when we intentionally reduce our operations and support capacity.
+See [](time-off:annual-expected) for our broader policies and support commitments during this time.
+
+(support:expected-time-off:policy)=
+### Expected time off policy
+
+Below is our policy for support during expected time off:
+
+- We monitor our support e-mail for major incidents, but do not guarantee a response for non-essential requests or questions.
+- We will take steps to minimize harm and avoid catastrophic problems (e.g. incidents that drastically increase costs), but will not perform non-essential actions. We will not treat "a general outage" as catastrophic unless we get an explicit signal in advance from a community representative they are doing essential work during this time.
```
GeorgianaElena, line 247: How much time does it mean "in advance" here?

GeorgianaElena:
> It implements the suggestions that y'all approved above, so if I don't hear any negative feedback or suggestions I'll assume we're all still in approval.

I generally agree with the approach here. Though, as lightweight this might be, this is still a small source of stress and a full disconnect from work won't be possible. I just want to acknowledge this so that we can continue to work on improving this for the next "generalized" expected time of period (when more than half of the team is expected to be AFK).

projects/managed-hubs/support.md
choldgraf: I am not sure, I struggled to define this without getting into too much detail. What I'm trying to convey is the idea that in some cases, we might get a community that makes an explicit request because some really important thing is happening. I wanted to suggest that we are open to conversation about this if it is really a problem for them.

Maybe it would be simpler to just say "sorry but we still can't promise anything" for now?

GeorgianaElena: > makes an explicit request because some really important thing is happening

This makes sense. Maybe one approach could be ask for this request to come before everyone leaves on vacation and there's enough time for everyone to plan around this (maybe 1, 2 weeks before)?

> Maybe it would be simpler to just say "sorry but we still can't promise anything" for now?

This approach makes me nervous because I feel like this policy translates into:
    *"we shouldn't expect anything from the support team, but we can hope they do everything"*

Which (for me at least) produces more stress than knowing exactly what are my responsabilities. Not sure if this makes any sense.

damianavila: > (maybe 1, 2 weeks before)?

2 weeks at minimum, I would say...

> Which (for me at least) produces more stress than knowing exactly what are my responsabilities. Not sure if this makes any sense.

I think the responsibilities during this time are generally well outlined in the above list. I do think we need to change our mindset to accommodate the vision on that list (I know it is not easy!!).

damianavila:
> I generally agree with the approach here. Though, as lightweight this might be, this is still a small source of stress and a full disconnect from work won't be possible. I just want to acknowledge this so that we can continue to work on improving this for the next "generalized" expected time of period (when more than half of the team is expected to be AFK).

I know... I think without a dedicated support team, the stress will be always there, so it is a matter to be pragmatic and reduce the stress as long as we have and I think this PR goes in that direction. Also, as I mentioned before, we need to make sure the load for these general time-off events is evenly distributed, so if you rotate in support on this iteration (ie. this EOY), you do rotate in the next one (ie. next EOY).

@GeorgianaElena, also notice my message here: https://github.com/2i2c-org/meta/issues/435#issuecomment-1341128899 that might help to reduce your stress in the upcoming rotation.

choldgraf:
I've pushed another edit to try and clarify that we _policy_ is "only absolutely essential work during holidays". I've also added a comment to note that in the future we may want to clarify what happens if a community strongly requests that we serve them over a holiday. But I don't think we need to decide that right now.

I agree with @GeorgianaElena's concerns in this issue:

- https://github.com/2i2c-org/meta/issues/435

But I am not sure if we can encode this in "2i2c policy". This feels like a cultural and team practices thing right now, so my hope is that the comment in the latest commit recognizes the possibility that others might request support nonetheless, but that we don't yet know the right _policy_ to use to handle this case.

So I recommend we close that issue with this PR and cross-link it (as I have done in the comment) to remember this topic.

Does that seem like an OK compromise for now? I think we can also use the conversation in this PR and in https://github.com/2i2c-org/meta/issues/435 to remember our concerns about this.

GeorgianaElena:
> Does that seem like an OK compromise for now?

+1 from me! Let's merge this 🚀 Thank you for all the discussions that happened here!

damianavila:
Indeed, let's get this one merged, I think it is good enough for now.

damianavila:
> So I recommend we close that issue with this PR and cross-link it (as I have done in the comment) to remember our conversation on this topic if we want to define more policy in the future.

Are you planning to keep https://github.com/2i2c-org/meta/issues/435 still open?

choldgraf:
if @GeorgianaElena is OK with it, I'd vote for closing that and re-opening if we want to bring the discussion back up (it is linked in the docs now)

GeorgianaElena:
Closed

PR #583, opened 2022-12-05
Add airtable information

choldgraf:
I've **upgraded our AirTable plan to the pro plan**. This will let us synchronize multiple airtables with the data that CS&S has provided, which we need for tracking invoices and contracts.

As a result, I've created a single shared AirTable account at `admin@2i2c.org`, and a password that I can share privately with anybody. We can work out a longer-term solution once we get a feel for it.

To begin, I've shared the password with @jmunroe but please let me know if you'd like access.

### DIFF
--- a/administration/airtable.md
+++ b/administration/airtable.md
@@ -1,0 +1,32 @@
+# AirTable accounts
+
+We use AirTable to track some of our leads, contracts, and invoicing data.
+This page covers this account and how to access it.
+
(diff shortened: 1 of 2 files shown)


damianavila (approved):

choldgraf:
Thanks Damian - I'll merge this one and we can iterate on subsequent changes to the setup.

PR #603, opened 2023-01-10
Document design location

choldgraf:
I just realized that our figma board wasn't documented, it is now!


sgibson91 (approved):

choldgraf:
thanks @sgibson91 !

PR #620, opened 2023-01-24
Document ramp receipt e-mail

choldgraf:
This documents a useful feature in Ramp to auto-detect receipts in e-mails.


damianavila (approved):

choldgraf:
@damianavila could you just merge these in the future instead of approving but not merging? I think the "let the PR author merge" practice makes sense for running infrastructure, but for most other things I think we should just merge as soon as somebody thinks it is reasonable (unless it's a complicated proposal or big change to how we work)

PR #622, opened 2023-01-25
Move documentation structure guidelines from infrastructure repo

sgibson91:
This page used to be in the infrastructure repo, but I think it makes more sense in our team compass. See related:

- https://github.com/2i2c-org/infrastructure/pull/2091


sgibson91:
Some weird git stuff going on there...

PR #626, opened 2023-01-26
Add a table of Support Stewards to our docs

### LINKED ISSUE #494: Auto-populate our docs with who our Meeting Facilitators/Support Stewards are from Slack usergroups
> This is a cool idea - we could even auto populate some headshots to show the latest members of each group in our team compass or something. 👍

_Originally posted by @choldgraf in https://github.com/2i2c-org/team-compass/issues/475#issuecomment-1185723865_

### Context

I created specific Slack usergroups to track who is serving in Support Steward/Meeting Facilitators role to (i) decouple the roles from each other a bit, and (ii) decouple the roles from membership of the `tech-team` usergroup. This is mostly so the code in team-roles-geekbot-sweep can handle this fine-grained approach as well.

However with these new usergroups, we can also automatically populate a list of people who serve in these roles for our docs. And perhaps pull people's Slack profile pictures. Hence creating a page that looks like this: https://jupyterhub-team-compass.readthedocs.io/en/latest/team/index.html

sgibson91:
This PR adds a table of all the team members who can serve as support stewards to our documentation. Names and avatars are pulled from Slack and appearance on this table is determined by membership of the `@support-stewards` Slack usergroup.

I added a Slack API token as an environment variable to the RTD project. Step 4 in this part of the documentation confused me: https://docs.readthedocs.io/en/stable/environment-variables.html#user-defined-environment-variables As I result, I have left the environment variable as "Private". This means that the table is not rendered in previews of the site in PRs, as we can see here: https://2i2c-team-compass--626.org.readthedocs.build/en/626/projects/managed-hubs/support.html#who-are-our-support-team But 🤞🏻 it should be when we merge. Here is a screenshot of the table from my local build:

Screenshot 2023-01-25 at 16 55 18

fixes #494

### DIFF
--- a/_data/support_stewards/gen_support_stewards.py
+++ b/_data/support_stewards/gen_support_stewards.py
@@ -12,7 +12,7 @@ class SlackUsergroupMembers:
     def __init__(self):
         token = os.environ.get("SLACK_BOT_TOKEN", None)
         if token is None:
-            raise ValueError("SLACK_BOT_TOKEN must be provided!")
+            print("SLACK_BOT_TOKEN does not exist. Skipping Support Steward table generation...")
 
         self.client = WebClient(token=token)
 
--- a/_data/support_stewards/gen_support_stewards.py
+++ b/_data/support_stewards/gen_support_stewards.py
@@ -1,0 +1,136 @@
+import os
+from collections import OrderedDict
+from pathlib import Path
+from textwrap import dedent, indent
+
+from slack_sdk import WebClient
+
+
+class SlackUsergroupMembers:
+    """Find the members of a given Slack usergroup"""
+
+    def __init__(self):
+        token = os.environ.get("SLACK_BOT_TOKEN", None)
+        if token is None:
+            raise ValueError("SLACK_BOT_TOKEN must be provided!")
+
+        self.client = WebClient(token=token)
+
+    def _get_usergroup_id(self, usergroup_name):
+        """Retrieve the ID of a named Slack usergroup
+        
+        Args:
+            usergroup_name (str): The display name of the desired Slack
+                usergroup
+        """
+        # List all usergroups in the workspace
+        response = self.client.api_call(api_method="usergroups.list")
+
+        # Find ID for the given usergroup
+        index = next(
+            idx
+            for (idx, usergroup) in enumerate(response["usergroups"])
+            if usergroup["handle"] == usergroup_name
+        )
+
+        self.usergroup_id = response["usergroups"][index]["id"]
+
+    def _get_user_ids(self, usergroup_name):
+        """
+        Retrieve the user IDs of the members of a given usergroup
+        """
+        self._get_usergroup_id(usergroup_name)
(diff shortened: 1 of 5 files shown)


projects/managed-hubs/support.md
```diff
@@ -107,6 +107,14 @@ Hub Administrators
   
 ```
 
+### Who are our Support Team?
+
+The Support Steward role rotates through the below Support Team members.
+
+```{eval-rst}
```
choldgraf, line 114: Not a big deal but why not just use MyST markdown for these instead of eval-rst ?

choldgraf (approved):
+1 on the feature from me (i think that slack group is a source of truth , right?)

I suggest that if nobody objects, Sarah just merges this and can self merge a few PRs to get it working on RTD if necessary

projects/managed-hubs/support.md
sgibson91: I was copying from the JupyterHub team compass docs, so it wasn't so much a "decision I made" as a "I have no idea how to do this so I'll copy an example I know works" scenario. I don't have enough working knowledge of MyST markdown to make on the fly replacements yet.

sgibson91:
> i think that slack group is a source of truth , right?

Yes, the Slack group is the source of truth for "who can serve as a support steward"

projects/managed-hubs/support.md
choldgraf: Mind if I push the equivalent to show how it is structured?

Also this makes me wonder - how can we expose Myst documentation to the team in a more accessible manner? We use MyST throughout our docs so I wanna make sure it isn't a barrier for others

sgibson91: Please feel free! I would love to be able to learn more, but right now I just have to go through others' examples and figure it out on the way.

choldgraf:
I just pushed a commit that I believe updates it to myst markdown. I also fixed the `conf.py` call because it was pointing to a file that didn't exist.

That said, one challenge I see with this implementation is that it will error if the slack API token doesn't exist locally...this might be a pain for people that don't have it in their environment, so maybe we should instead check if that token exists and print a message like "{TOKEN} doesn't exist, skipping support steward generation..." rather than raising an error

_data/support_stewards/gen_support_stewards.py
```diff
@@ -0,0 +1,139 @@
...
+from slack_sdk import WebClient
+
+
+class SlackUsergroupMembers:
+    """Find the members of a given Slack usergroup"""
+
+    def __init__(self):
+        token = os.environ.get("SLACK_BOT_TOKEN", None)
+        if token is None:
+            raise ValueError("SLACK_BOT_TOKEN must be provided!")
```
sgibson91, line 15:
```suggestion
            print("SLACK_BOT_TOKEN does not exist. Skipping Support Steward table generation...")
```

sgibson91:
I'm going to merge this and see if it works

PR #627, opened 2023-01-26
Fix support stewards card

choldgraf:
The markdown to generate the support stewards cards didn't have the proper nesting, so I think this fixes it.


sgibson91 (approved):
Yaaay, works now!

PR #628, opened 2023-01-26
Prevent autobuild from looping

sgibson91:
This redirects the output of the gen_support_stewards.py file to the tmp dir which is already both git- and sphinx-ignored. This prevents the live build from continuously looping because the _data/support_stewards folder always had changes due to regeneration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ tmp/
 .DS_Store
 .nox
 __pycache__
-_data/support_stewards/*.txt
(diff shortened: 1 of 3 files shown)


choldgraf (approved):
seems reasonable to me - we might need to edit the `noxfile.py` call to do `.resolve()` on each of the `AUTOBUILD_IGNORE` paths if this doesn't work

sgibson91:
It works locally for me

Issue #641, opened 2023-02-03
Broken List of Team Members page


jmunroe:
### Context

Our "List of Team Members" page on the Team Compass appears broken:

https://compass.2i2c.org/en/latest/reference/team.html

<img width="1127" alt="image" src="https://user-images.githubusercontent.com/6181563/216640402-d35a6957-06c5-48dd-850f-e8891839fa92.png">

choldgraf:
holy moly!

PR #643, opened 2023-02-03
Fix team page

choldgraf:
closes https://github.com/2i2c-org/team-compass/issues/641


jmunroe:
Thanks. That should fix it!

jmunroe (approved):
LGTM

PR #646, opened 2023-02-03
Add invoicing procedures

choldgraf:
This adds some basic instructions for sending invoices to CS&S, including the admin addresses to cc on the invoice.

cc @consideRatio - would this have been enough information for you? Or perhaps we should link out to the CS&S invoicing docs too?

### DIFF
--- a/administration/index.md
+++ b/administration/index.md
@@ -7,6 +7,7 @@ This chapter contains administrative information at 2i2c.
 structure
 css
 reimburse
+invoices
 google-workspace
 github
 airtable
--- a/administration/invoices.md
+++ b/administration/invoices.md
@@ -1,0 +1,6 @@
+# Send invoices
+
+For any team members that are paid off of fixed-term contracts, they must send monthly invoices to receive payment for their efforts.
+This page describes the process for doing so.
+
+1. **Prepare an invoice template**. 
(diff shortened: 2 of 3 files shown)


consideRatio (approved):
- CS&S have documentation on how to provide invoices that is worth linking out to.
- The grant code. Who decides what code to use / who to contact about it in 2i2c, and/or where to read about them? It could be good if there is a link to help 2i2c admins making such call to find their way to such document directly from here, if there is a grant code document to link to or similar.

damianavila:
> The grant code. Who decides what code to use / who to contact about it in 2i2c, and/or where to read about them? It could be good if there is a link to help 2i2c admins making such call to find their way to such document directly from here, if there is a grant code document to link to or similar.

This is a very good point raised by @consideRatio.

choldgraf:
We actually ran into the same issue over in this PR:

- https://github.com/2i2c-org/team-compass/pull/659

So I've added a link to these docs that cross-references the "how to find my grant code" section. I'll merge this one since I think @consideRatio 's comments are now addressed!

damianavila (approved):

choldgraf:
One sec actually, I'm gonna move some of the "invoicing instructions" docs that were in reimbursements back into "invoices" and cross-link from there. That's a more natural space I think

choldgraf:
ok now done + merged!

consideRatio:
Wieee thanks @choldgraf!!

PR #657, opened 2023-02-10
Add mastodon account

choldgraf:
Documents our mastodon account. Also replaces "tweet" with "post to social media". While I was at it, I realized that we had two different blog post checklists (way to go Chris). So this also unifies them into a single issue template.

(also fixes up some whitespace bug that was in the partnerships page)


damianavila (approved):

choldgraf:
I'm going to merge this one in, since it documents a pre-existing thing, but I'm happy to iterate on the content or make updates to the hachyderm page if people would like!

PR #659, opened 2023-02-13
Update reimbursement docs

choldgraf:
This updates our reimbursement docs to make it clear how employees vs. contractors can reimburse


pnasrat (approved):
This makes the process much clearer. 

Could probably also create a cover template google doc for CS&S employee reimbursement, but that can be a later addition.

LGTM

choldgraf:
Thanks - and agreed re: employee reimbursement

Issue #666, opened 2023-02-15
Switch our documentation to use `dirhtml` instead of `html`


choldgraf:
Currently we are using the default Sphinx builder for `html` outputs, which creates an individual file with `.html` at the end. For example:

https://compass.2i2c.org/en/latest/open-source/strategy.html

However most modern website builders default to using a **directory** as the build output, with an `index.html` file inside, so that instead the link structure becomes:

https://compass.2i2c.org/en/latest/open-source/strategy

I think this structure is much more common in general these days, and most other SSGs use this structure (including 2i2c.org). It is also easily do-able with Sphinx, via `-b dirhtml`.

I propose that we migrate all of our documentation to use the `dirhtml` builder and update links accordingly.

### Sites we'd need to update

- docs.2i2c.org
- infrastructure.2i2c.org
- compass.2i2c.org

consideRatio:
Do you know if this change break existing links in some way? For example, will https://compass.2i2c.org/en/latest/open-source/strategy.html stop working when https://compass.2i2c.org/en/latest/open-source/strategy starts working?

choldgraf:
yep it will, I think we'll need to do a `ctrl+f` for any `.html` links in our docs and remove it. That shouldn't be too hard as long as it's linked from within `2i2c-org`. E.g. here's a search we can use:

https://github.com/search?q=org%3A2i2c-org+%22.html%29%22&type=code

For any document / label linked directly by Sphinx, it will automatically change, we don't need to change that.

consideRatio:
The key question in my head then is if we accept breaking all external existing links to accomplish this, and/or if we can use rediraffe to help redirect the `something.html` versions to `something/` instead.

choldgraf:
I don't think that rediraffe would work, because it operates at the document level rather than at the end-link level unfortunately :-/

damianavila:
The proposed migration makes sense to me.

choldgraf:
### We could manually define the pages to re-direct

I briefly looked into this - `sphinx-rediraffe` doesn't let you redirect from something like `mypage.html` to `mypage/index.html`. However, the template they are using is pretty simple:

https://github.com/wpilibsuite/sphinxext-rediraffe/blob/d8961c2402015dafa6039ed27d973ad835d7fc09/sphinxext/rediraffe.py#L21-L39

<details><summary>Their template</summary>
<p>

```
DEFAULT_REDIRAFFE_TEMPLATE = Template(
    """
<html>
    <head>
[...]

PR #667, opened 2023-02-15
Change to dirhtml builder

choldgraf:
This updates to the `dirhtml` builder so that URLs no longer have `.html` at the end of them.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,6 +12,7 @@ build:
 
 sphinx:
   configuration: conf.py
+  builder: dirhtml
 
 python:
   install:
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To do so, follow these steps:
    $ nox -s docs
    ```
 
-This should create a local environment in a `.nox` folder, build the documentation (as specified in the `noxfile.py` configuration), and the output will be in `_build/html`.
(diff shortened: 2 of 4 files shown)


damianavila (approved):
LGTM.

sgibson91:
Do we have any redirection setup on this repo so that links don't break when we merge this? I'm specifically thinking about cross-links to docs built in other repos.

choldgraf:
It probably wouldn't be straightforward, see some context here: https://github.com/2i2c-org/team-compass/issues/666#issuecomment-1431331029

We should also decide if we want to drop "en/latest" while we are at it...

Either way we can hold off on this until I'm back from pat leave and have time to do this all at once. I don't think it's urgent now

sgibson91:
Of course! I asked in case it was simple, because then I would've just merged the PR :)

choldgraf:
Ah indeed - I'll try to provide some more context in that issue so that it's clear what steps we'll need to take and it'll be more actionable once it's time to take action 🙂

choldgraf:
re-opening this PR to see if the `.html` -> `folder` redirection is now working in our theme! If it is then I will merge this

Tikhon88 (commented):
@vdatg

PR #673, opened 2023-02-17
Clarify travel reimbursement

choldgraf:
These add a few travel reimbursement guidelines so that we can disambiguate any confusion or questions ahead of time! It tries to be relatively lightweight but hopefully helpful. Comments or suggestions welcome.

### DIFF
--- a/administration/reimburse.md
+++ b/administration/reimburse.md
@@ -40,7 +40,7 @@ Below are a few guidelines for travel reimbursements with 2i2c:
 - We'll cover the cost of lodging for the day before or after.
 - For international travel, we can cover an extra day before to give more time to adjust.
 - It is fine if you are arriving / leaving from two different locations (e.g., if you intend on making a side-trip after the event).
-  In this case, 2i2c will only cover the trip to / from the vent, not the third leg.
+  In this case, 2i2c will only cover the trip to / from the event, not the third leg.
 
 And some general tips:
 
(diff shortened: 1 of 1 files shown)


administration/reimburse.md
```diff
@@ -26,10 +26,27 @@ Currently, these are the expenses that we regularly reimburse:
...
+(reimburse:travel)=
+### Travel reimbursements
+
+Below are a few guidelines for travel reimbursements with 2i2c:
+
+- 2i2c will cover the cost of travel to and from the event.
+- We'll cover the cost of lodging for the day before or after.
+- For international travel, we can cover an extra day before to give more time to adjust.
+- It is fine if you are arriving / leaving from two different locations (e.g., if you intend on making a side-trip after the event).
+  In this case, 2i2c will only cover the trip to / from the vent, not the third leg.
```
pnasrat, line 43: typo vent -> event

choldgraf, line 43:
```suggestion
  In this case, 2i2c will only cover the trip to / from the event, not the third leg.
```

damianavila (approved):
These are sensible guidelines, IMHO.

choldgraf:
Folks, please just merge these if you think they are reasonable and they don't have any objections! We don't need dangling PRs with approvals but no objections sitting around for too long :-)

Issue #655, opened 2023-02-10
[proposal] Support process to be led by Engineering


choldgraf:
Our [support process](https://compass.2i2c.org/en/latest/projects/managed-hubs/support.html) and the related [incident response process](https://compass.2i2c.org/en/latest/projects/managed-hubs/incidents.html) currently does not have a clear "owner" that is responsible for the process itself, and that has the authority to modify and change the process as needed.

In a recent [organization strategy meeting](https://docs.google.com/document/d/1HoNX8T8IQ1uhS2ryi1r9iS-nSbPT1b1Y7HsjosbHme8/edit#), there seemed to be agreement that we should formalize **engineering as overseeing our support systems**. They will need to gather feedback from other parts of the org, but they are the ones formally responsible the system and with the authority to change it.

### Rationale

We all agreed that support is inherently cross-organization, and also agreed that it is not efficient to have engineers doing direct-line support all the time. However, we should define a policy that reflects our current reality first, and then we can shoot for taking steps that improve that policy / system.

In the short term, the engineering team feels the most impact from our support system. They are the ones that are handling most of the requests, and the system for how those requests are managed will impact their well-being more than others. Moreover, most support requests result in the need for action from engineering, and so they need to be a central part of designing the process.

### What does this mean?

I think that this means that engineering has the authority to design the system of roles and processes that defines "support at 2i2c", and to advocate for the resources needed to fill those roles (e.g. requesting help from others at 2i2c or signaling that we need to find resources to pay for more effort on this). Again this will naturally be done in collaboration with others in 2i2c, but engineering leads the effort and has an outsized role in the decision-making.

### In the future

We may want to revisit whether support should "live" inside engineering in the long term, or if it should instead be either a multi-area process by design, or live within another part of the organization. The goal of this is to define our current situation, and we should revisit this setup once 2i2c has more structure and resources (e.g., if we have a cross-functional project manager that can help us coordinate actions across our areas).

choldgraf:
cc'ing the area leads/project managers for visibility, though I think we've already discussed this one and have agreement so unless somebody _objects_, we can probably move forward with this.

damianavila:
I presume this is the section where we should make the clarification in our team-compass: https://github.com/2i2c-org/team-compass/blob/main/projects/managed-hubs/support.md#roles-and-team-structure. Any other suggestions?

pnasrat:
Related to this I've created an issue to track process improvements and definition for the support process in #674

PR #676, opened 2023-02-23
Support process lead by Engineering.

damianavila:
This change actually documents in our team compass the agreement we achieved as described in https://github.com/2i2c-org/team-compass/issues/655.

### DIFF
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -9,6 +9,8 @@ The content on this page might change over time, and we welcome suggested change
 
 This section contains information that our team uses to triage, communicate, and resolve {term}`Support Requests` that come from communities.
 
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
LGTM, thank you @damianavila!!

I think this is agreed on elsewhere and it makes sense to go for a quick merge!

choldgraf:
A quick thought here - I think that it'd be helpful if we kept the "scope and responsibilities" section of each area as a single source of truth for what they're responsible for. So I suggest that we add a short section there for engineering, and it can just delegate + link to these support docs for more information.

damianavila:
I agree, I will do that once the big refactor PR lands so you don't have to deal with conflicts on that one.

damianavila:
> I agree, I will do that once the big refactor PR lands so you don't have to deal with conflicts on that one.

See https://github.com/2i2c-org/team-compass/pull/680.

PR #681, opened 2023-03-08
Add a targeted hiring section

damianavila:
Closes https://github.com/2i2c-org/meta/issues/423.

cc @sgibson91 who was interested to surface in our docs the EDI perspective this targeted hiring process should always have.


sgibson91 (approved):
:+1:

Issue #688, opened 2023-03-27
Update our financial KPIs board with the new CS&S structure


choldgraf:
CS&S updated their monthly account data that they've provided us. We need to update our KPIs board to accommodate this new data structure, and also update our team compass to reflect the new data that we get.

choldgraf:
This PR is an attempt at fixing our KPIs:

- https://github.com/2i2c-org/kpis/pull/14

choldgraf:
OK it looks like the KPIs work again! I also opened up this PR to update our docs, which should close this:

- https://github.com/2i2c-org/team-compass/pull/689

PR #689, opened 2023-03-27
Update our accounting table information

choldgraf:
This updates our accounting table and airtable information to reflect the latest changes from CS&S. It also describes the different AirTable bases in a bit more detail so that it's easier to find them.

- closes https://github.com/2i2c-org/team-compass/issues/688


choldgraf:
I'm going to merge this one so that I can iterate on it in a new PR - it's not adding any new information, just updating for new data structures from CS&S.

PR #690, opened 2023-03-29
Add clarifying message about budget projections

choldgraf:
In a recent meeting, @jmunroe noted that it was unclear whether others were encouraged to make **edits** to our budgeting projections sheet, or if it was meant to be a static thing that only I updated. This tries to clarify that yes, others should feel free to edit if it is helpful for making decisions.

- Original page: https://compass.2i2c.org/en/latest/finance/projections.html


GeorgianaElena (approved):

damianavila (approved):

damianavila:
Merged since we have at least 2 approval and I do not think there is controversy at all.

Issue #496, opened 2022-08-26
Re-work our team meeting structure to improve cross-team communication


choldgraf:
### Context

When 2i2c was first created, we had a single engineering team focused around our infrastructure and services around it. Now, we have at least two teams - an engineering team (@2i2c-org/tech-team) and a more community-facing team (@2i2c-org/partnerships-and-community-guidance).

Other than just generally talking about stuff on GitHub and Slack, there are two places where we intentionally get together to discuss:

- A bi-weekly planning meeting - this is a very tactical meeting that is largely about telling people what we're up to.
- A monthly all-team meeting - this is a very high-level meeting where we tend to zoom way out

However, I believe this system is not properly facilitating communication, context-sharing, and collaboration between these two teams.

It is crucial that 2i2c builds a team system that allows its different components to align and move in-step with one another. This will ensure that the knowledge present in each team is used to guide the decisions of the other team.

### Proposal

We should propose some iterative changes to make that will improve the communication and coordination between @2i2c-org/tech-team and @2i2c-org/partnerships-and-community-guidance . I invite others to share ideas below. A few questions that we can try to answer:

- Do we need another, or a different, meeting for this?
- Would it require an "all hands" kind of meeting or can we do this with sub-groups of people?
- Is there an asynchronous way to encourage this more effectively?
- Who is the person that should "oversee" these kinds of inter-team communication meetings to make sure they happen? Is there a role that we are missing? Is this the Project Manager or is that out of scope?

```[tasklist]
- [x] https://github.com/2i2c-org/team-compass/issues/518
- [ ] https://github.com/2i2c-org/team-compass/issues/517
- [ ] https://github.com/2i2c-org/team-compass/issues/737#top
```

choldgraf:
A few ideas that others had shared in a brief conversation about this on Slack:

- @sgibson91: maybe we could use one of those "random pairing for coffee" generators as well? https://www.donut.com/ 
- @sgibson91: Maybe a regular team meeting that's broader across 2i2c as well? Currently the meeting is very engineering heavy (though Jim has attended in the past!)
- @sgibson91: a regular in-person catch-up (6 monthly?) would be lovely
- @colliand: Let's all move into a castle in Romania! (😅)

damianavila:
[...]

PR #691, opened 2023-03-29
Be more specific about agendas for meetings

yuvipanda:
- Specify the amount of time in hours, rather than the more nebulous 'working day'. I think 36h without counting weekends should count for most TZ drift
- Provide slightly more detail on what shape the agenda should be in.

### DIFF
--- a/operations/meetings.md
+++ b/operations/meetings.md
@@ -17,16 +17,20 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
   These must be made available to the team reasonably soon after the meeting.
 
 - **Meeting agendas must be available at least 36 hours (not counting weekends) ahead**.
-  Agendas should:
-  1. List any specific **decisions** people want to make progress on during the meeting.
-  2. Provide ideas for topics of possible discussion.
+  Agendas should list:
+  1. Any **questions** that people wish to discuss and answer during the meeting.
+  2. Any **decisions** people want to make progress on during the meeting, and where the final decision will be made.
+  3. Any **topics** to discuss in the meeting, with enough detail to help others decide if they wish to participate.
 
   Agendas should be concrete enough that people who can not be synchronously
   present can make comments on it, as well as make choices on whether they
   would like to attend or not.
+  Provide links to the **source of truth for discussion**, as well as any context needed to meaningfully participate.
 
-  If a meeting does not have an agenda 36 hours (not counting weekends) in
-  advance, then team members may assume it is cancelled and rescheduled.
+  **If there's not an agenda 36 hours before, assume the meeting is cancelled**.
+  Meeting facilitators are responsible for collecting potential agenda items from others ahead of time, and setting the meeting agenda 36 hours prior to the meeting start.
+  If there is no agenda at this point, the meeting should be deemed as cancelled.
+  The meeting facilitator should inform the team and delete the calendar event.
 
 - **Focus time on resolving uncertainty and making decisions**.
   Avoid the tendency to use meetings for updates - we should use asynchronous team processes (discussion in issues, Geekbot standups, etc) for sharing what we are up to day-to-day.
--- a/operations/meetings.md
+++ b/operations/meetings.md
@@ -22,7 +22,7 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
   2. Provide ideas for topics of possible discussion.
 
   Agendas should be concrete enough that people who can not be synchronously
-  present can make comments on it, as well as make choices on wether they
+  present can make comments on it, as well as make choices on whether they
   would like to attend or not.
 
   If a meeting does not have an agenda 36 hours (not counting weekends) in
--- a/operations/meetings.md
+++ b/operations/meetings.md
@@ -15,9 +15,19 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
 - **Take notes and make them accessible**.
   Meetings must have written notes that are accessible to others on the team.
   These must be made available to the team reasonably soon after the meeting.
-- **Meeting agendas must be available at least one working day ahead**.
-  Agendas should provide an idea for the topics of discussion and any major questions to resolve.
-  If a meeting does not have an agenda at least one working day in advance, then team members may assume it is cancelled.
+
+- **Meeting agendas must be available at least 36 hours (not counting weekends) ahead**.
+  Agendas should:
+  1. List any specific **decisions** people want to make progress on during the meeting.
+  2. Provide ideas for topics of possible discussion.
+
+  Agendas should be concrete enough that people who can not be synchronously
+  present can make comments on it, as well as make choices on wether they
+  would like to attend or not.
+
+  If a meeting does not have an agenda 36 hours (not counting weekends) in
+  advance, then team members may assume it is cancelled and rescheduled.
+
 - **Focus time on resolving uncertainty and making decisions**.
   Avoid the tendency to use meetings for updates - we should use asynchronous team processes (discussion in issues, Geekbot standups, etc) for sharing what we are up to day-to-day.
 
--- a/operations/meetings.md
+++ b/operations/meetings.md
@@ -27,10 +27,9 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
   would like to attend or not.
   Provide links to the **source of truth for discussion**, as well as any context needed to meaningfully participate.
 
-  **If there's not an agenda 36 hours before, assume the meeting is cancelled**.
-  Meeting facilitators are responsible for collecting potential agenda items from others ahead of time, and setting the meeting agenda 36 hours prior to the meeting start.
-  If there is no agenda at this point, the meeting should be deemed as cancelled.
-  The meeting facilitator should inform the team and delete the calendar event.
+- **If there's not an agenda 36 hours before, the facilitator should cancel the meeting**.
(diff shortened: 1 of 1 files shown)


pnasrat:
Perhaps add if an agenda item is coming to a decision on something or discussing detail - a design, etc, there should be a source of truth (github issue, design doc, etc) linked to the agenda item (and note if that's the place for feedback). Also reminding people not to assume shared context (eg provide links or a section on  background and reasoning)

pnasrat:
However I won't block on either of those as this is definitely an improvement

yuvipanda:
@pnasrat steal something from https://theapacheway.com/on-list/ perhaps?

GeorgianaElena (approved):
Thank you for updating this  @yuvipanda! 
I really like @pnasrat's proposed additional ideas, but also agree it shouldn't be a blocker as this is already a very good step forward!

operations/meetings.md
```diff
@@ -15,9 +15,19 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
...
-  Agendas should provide an idea for the topics of discussion and any major questions to resolve.
-  If a meeting does not have an agenda at least one working day in advance, then team members may assume it is cancelled.
+
+- **Meeting agendas must be available at least 36 hours (not counting weekends) ahead**.
+  Agendas should:
+  1. List any specific **decisions** people want to make progress on during the meeting.
+  2. Provide ideas for topics of possible discussion.
+
+  Agendas should be concrete enough that people who can not be synchronously
+  present can make comments on it, as well as make choices on wether they
```
sgibson91, line 25:
```suggestion
  present can make comments on it, as well as make choices on whether they
```

yuvipanda: wether it is whether or wether, but maybe it's weather? hmm, perhaps wether is not actually a word :D

sgibson91: My friend thought a "sweater" was actually "sweather" for the longest time and it was too funny to correct

yuvipanda: sweeter sweather weather

jmunroe:
I think there has to be a middle ground between 'discussions' and 'decisions'.  I fully support that decisions raised in a meeting should be put on an agenda with sufficient detail and time so that async input is possible.  That might mean that a 'decision' formulated during a group meeting needs to go through a process before it is adopted to give time for input of others.

However, I am concerned about requiring 'discussions' to have sufficient detail to participate in prior to the meeting. For me, discussions are, often, brought to a group meeting because the idea is poorly formed and premature so need conversation to better formulated. I would be very challenged to give sufficient detail 36 hrs before a meeting on a discussion that hadn't happened yet.  

I agree async input is important. Can we build up **automated** tooling of 1) video recording team meeting and posting the link 2) creating a searching transcript 3) using one of the many AI tools out there to create 'meeting summaries' in addition to 4) notes take in the meeting?

yuvipanda:
> However, I am concerned about requiring 'discussions' to have sufficient detail to participate in prior to the meeting.

I think we should clarify the wording so that the idea is that *what* discussions are planned is what is specified, rather than having the discussions themselves 36 hours before.

As an example, here's an item that I added to the agenda for Tuesday:

> Discuss https://github.com/2i2c-org/team-compass/pull/691 for better asynchronous meetings
> - Chat about a light-weight RfC process in the medium-term

I think that's sufficient detail for what is a 'discuss about X' topic.

jmunroe:
Thanks @pnasrat . That's useful to know what level of detail would be helpful.  Grants/leads/projects always (should) have at least a placeholder github issue and/or gdrive folder link.

choldgraf:
To me the important questions people need to be able to answer ahead of time are:

- Is this a meeting I want to attend in person?
- Where can I find the context needed to meaningfully participate?

And also that for those that don't show up in person:

- they can meaningfully participate asynchronously
- and with the confidence that all relevant points in the meeting are recorded in an easy to discover place

I'm a big fan of anything that gets us closer to this goal. This seems like a good step. Maybe we should document some of the clarifying points people shared above to make it clearer what context is needed.

Another thing we might consider is being explicit about when we **expect** attendees to read, digest, etc *before* the meeting in order to be able to have a productive discussion. This may not always be the case, but sometimes meetings benefit from a pre-read so you can jump straight into discussion.

damianavila:
I would encourage taking a step-by-step approach (incremental changes) to this practice to make sure we do not end up with a long process that nobody follows. So, my proposal is to merge this one as is in the next 24hs (unless somebody else feels strongly opposed), practice the guidelines for a few weeks, and revisit again to make further improvements.

damianavila (approved):

operations/meetings.md
```diff
@@ -15,9 +15,19 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
 - **Take notes and make them accessible**.
   Meetings must have written notes that are accessible to others on the team.
   These must be made available to the team reasonably soon after the meeting.
-- **Meeting agendas must be available at least one working day ahead**.
-  Agendas should provide an idea for the topics of discussion and any major questions to resolve.
-  If a meeting does not have an agenda at least one working day in advance, then team members may assume it is cancelled.
+
+- **Meeting agendas must be available at least 36 hours (not counting weekends) ahead**.
+  Agendas should:
+  1. List any specific **decisions** people want to make progress on during the meeting.
```
choldgraf, line 21:
```suggestion
  1. Any **questions** that people wish to discuss and answer during the meeting.
  2. Any **decisions** people want to make progress on during the meeting, and where the final decision will be made.
```

operations/meetings.md
```diff
@@ -15,9 +15,19 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
...
+  2. Provide ideas for topics of possible discussion.
+
+  Agendas should be concrete enough that people who can not be synchronously
+  present can make comments on it, as well as make choices on whether they
+  would like to attend or not.
+
+  If a meeting does not have an agenda 36 hours (not counting weekends) in
+  advance, then team members may assume it is cancelled and rescheduled.
```
choldgraf, line 29:
```suggestion
  **If there's not an agenda 36 hours before, assume the meeting is cancelled**.
  Meeting facilitators are responsible for collecting potential agenda items from others ahead of time, and setting the meeting agenda 36 hours prior to the meeting start.
  If there is no agenda at this point, the meeting should be deemed as cancelled.
  The meeting facilitator should inform the team and delete the calendar event.
```

damianavila:
> I have a few suggestions for improving the wording here, given the discussion in this PR. I don't think it meaningfully changes the intent, just clarifies things a bit more. What do you think?

I like all the suggestions made.

choldgraf:
I have two other suggestions to add:

1. **What if we moved this to 24 hours instead of 36 hours**? I find that a lot of movement around topics that we might discuss happens in the day before a meeting, and I end up setting the meeting agenda about 1 day ahead of the meeting time. If this were 36 hours, then it is effectively 48 hours unless I set the meeting time in the middle of the night. I foresee this will already be a practice that we have a hard time meeting and it seems 24 hours is more realistic to me.
2. **Ask meeting facilitators to explicitly cancel the meeting or share the agenda ahead of time**. I often see a meeting on the calendar a few hours from now, and then look at the agenda to discover that there's nothing there. I think we should have a decision point at the "share an agenda" moment where the facilitator explicitly calls the meeting off (and deletes the calendar event) **or** shares the agenda link with the team.

Any feedback on that? My hope is that this will help team facilitators be more accountable to putting up well-structured meeting agendas for others to participate in.

yuvipanda:
@choldgraf +1 on having the facilitator explicitly cancel the meeting. This also means all meetings *must* have an explicit facilitator.

I picked 36h to kinda be like '24 but be TZ inclusive'. Ideally I'd like someone to make a decision *today* about what meetings they want to go to *tomorrow*, and given how TZ diverse we are, 24 feels like it's not enough. I picked 36 as a compromise here, but can do 48 too.

jmunroe:
It looks like #496 might the best place to continue the overall conversation. I'll place my comments there.

yuvipanda:
@choldgraf how do you feel about merging this PR as is now, and we can discuss the time situation separately?

operations/meetings.md
```diff
@@ -15,9 +15,23 @@ As a distributed team, we pay a high coordination penalty for getting multiple p
...
+
+  Agendas should be concrete enough that people who can not be synchronously
+  present can make comments on it, as well as make choices on whether they
+  would like to attend or not.
+  Provide links to the **source of truth for discussion**, as well as any context needed to meaningfully participate.
+
+  **If there's not an agenda 36 hours before, assume the meeting is cancelled**.
+  Meeting facilitators are responsible for collecting potential agenda items from others ahead of time, and setting the meeting agenda 36 hours prior to the meeting start.
+  If there is no agenda at this point, the meeting should be deemed as cancelled.
+  The meeting facilitator should inform the team and delete the calendar event.
```
choldgraf, line 33:
```suggestion
- **If there's not an agenda 36 hours before, the facilitator should cancel the meeting**.
  Meeting facilitators should collect _potential_ agenda items from others ahead of time, and _set_ the agenda at least 36 hours prior to the meeting start.
  If there is no set agenda at this time, facilitator should inform the team in Slack and delete the calendar event.
```

choldgraf: how about this as an iterative improvement and we can discuss others in another thread / in the team meeting?

PR #698, opened 2023-04-04
Document our Invoices and Contracts airtable

choldgraf:
This PR updates our documentation regarding invoices and contracts, and cleans up some more of our accounting and sales information.

Major changes:

- Documents two synced AirTable tables that we now pull directly from CS&S.
- Documents an "interface" (dashboard) that we use to summarize this AirTable data
- Removes most of our "leads and sales pipeline" information because it is out of date. (@colliand please update this section with the latest process that you have developed.)


choldgraf:
@colliand can you please review this and let me know if the explanations there are clear? I think it's important that you have a strong understanding / regular interaction with these tables moving forward so make sure that it makes sense to you.

@pnasrat this is the AirTable data I was describing in Slack, in case you'd like to take a look and provide feedback. In particular I'm curious if you think this is a reasonable "source of truth" for this data for the purposes of https://github.com/2i2c-org/infrastructure/issues/2431

colliand:
I will review and report back.

choldgraf:
Quick update: Joe at CS&S said that they can get us the "raw data" accounting information so we can use that as well for our airtable once they expose it.

colliand (approved):
These are excellent improvements to our documentation on single-source-of-truth usage of data from CS&S via AirTable. Thanks @choldgraf!

I note that the text on the sales process needs to be refactored (by @colliand) and merged with work done earlier this year in the Partnerships section of the team-compass.

PR #701, opened 2023-04-06
Clarify credit card numbers

choldgraf:
In Slack @yuvipanda asked a question about whether a credit card associated with AWS was 2i2c's or my personal card. This PR clarifies our two credit cards and adds the final four digits of each for reference.


yuvipanda (approved):
very helpful!

PR #710, opened 2023-04-12
Convert links to use dirhtml

choldgraf:
This makes our link builder use directories as outputs instead of html files. So instead of `mysite.org/page.html`, we'll be able to use `mysite.org/page`.

This is a copy of #667 since I think RTD was getting the dependency install wrong. If the redirects


choldgraf:
fixed by https://github.com/2i2c-org/team-compass/pull/667 in the end

PR #714, opened 2023-04-13
Document authentication accounts

choldgraf:
This documents the authentication services that we use for our hubs. Primarily it documents CILogon. It also mentions `Auth0` but states that we no longer user it. It also removes auth0 from our onboarding steps.

- closes https://github.com/2i2c-org/infrastructure/issues/2422


GeorgianaElena (approved):
Thanks @choldgraf

PR #727, opened 2023-04-27
Clarify airtable access

choldgraf:
I noticed that @damianavila asked for and received write access to the AirTable. We currently are keeping the AirTable to one account with edit permissions, and share that account via BitWarden. Otherwise, each person we add with editing permissions will incur an extra $22/month charge, and at least right now I don't see the value in having many people with separate editing privileges on top of a shared account.

This PR tries to clarify this!

### DIFF
--- a/administration/airtable.md
+++ b/administration/airtable.md
@@ -4,19 +4,20 @@
 We use AirTable to track some of our leads, contracts, and invoicing data.
 This page covers this account and how to access it.
 
-[Here's a link to our primary AirTable workspace](https://airtable.com/appbjBTRIbgRiElkr).
-In addition, there are several bases within this workspace that are described below.
+## Read-only access to the AirTable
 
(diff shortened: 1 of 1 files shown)


damianavila (approved):
Thanks for the clarification, @choldgraf!

damianavila:
> I noticed that @damianavila asked for and received write access to the AirTable. 

Should you remove the write access I now have so we can save those 22 USD/month charge, @choldgraf?

choldgraf:
Thanks @damianavila - and yep, that should now be done

damianavila:
> and yep, that should now be done

Is that tracked somewhere, if not can you confirm that happened? Otherwise, we are going to forget it.

choldgraf:
Already completed 👍

damianavila:
Thanks for the confirmation!!

PR #734, opened 2023-04-28
Add some info about invoicing expenses in other currencies

### LINKED ISSUE #733: Reimbursements in different currencies?
### Context

I read through our team compass regarding reimbursements and ended up at CS&S's documentation [here](https://www.codeforsociety.org/resources/getting-paid-by-css), but my question is still unanswered so I'm opening an issue.

The context of my question is that a contractor has chosen to pay for something themselves and be reimbursed by adding it as a line item on their invoice. But the currency of the line item for the `$something` is different to the line item for our contract, which I believe is USD.

How does CS&S want us to handle that difference? Should I convert the cost of the `$something` into USD? But then the receipt won't match (which I'm assuming we'd have to submit as evidence), so would I also need documentation of the exchange rate on the date of purchase?

### Proposal

_No response_

### Updates and actions

_No response_

sgibson91:
fixes #733

### DIFF
--- a/administration/reimburse.md
+++ b/administration/reimburse.md
@@ -57,6 +57,8 @@ To do so, see [](invoices.md).
 ## Reimbursement for contractors
 
 To reimburse expenses as a contractor, **submit a separate line item on your monthly invoice**, called `Reimbursement - <item>`.
+If you paid the expense in a currency other than USD, please convert it to USD using the daily conversion rate.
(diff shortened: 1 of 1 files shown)


consideRatio (approved):

consideRatio (approved):
:tada: thanks for clarifying this!!

choldgraf (approved):

PR #743, opened 2023-05-25
Quick updates about credit cards and CS&S accounting

choldgraf:
Two small updates, one that notes the CS&S accounting team + contact, and another that notes that we can raise our credit card limit on an as-needed basis if need be.

### DIFF
--- a/finance/payments.md
+++ b/finance/payments.md
@@ -27,3 +27,11 @@ Ends in `7770`.
 
 Used only for paying cloud providers (GCP, AWS, and Azure) for cloud costs.
 We generally intend to re-imburse ourselves via invoices to communities for any cost charged to this credit card.
+
+### Extending card spending limits
+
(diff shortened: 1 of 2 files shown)


finance/payments.md
```diff
@@ -27,3 +27,11 @@ Ends in `7770`.
 
 Used only for paying cloud providers (GCP, AWS, and Azure) for cloud costs.
 We generally intend to re-imburse ourselves via invoices to communities for any cost charged to this credit card.
+
+### Extending card spending limits
+
+We have a monthly spending limit on each card, and credit card purchases will begin being declined if we hit it.
```
choldgraf, line 33:
```suggestion
We have a monthly spending limit on each card, and credit card purchases will be declined if we hit it.
```

choldgraf:
This one is just documenting a few pieces of information so I will self-merge

PR #750, opened 2023-07-06
Create job description for Technical Content Specialist

jmunroe:
I'm putting this initially in the community section. After the reorg of community into partnerships, we can move this role.


damianavila (approved):
I presume this is the result of the "live" job description, then I think it should be merged as is.

PR #764, opened 2023-08-16
Update non-incident support workflow

damianavila:
I want to share with you a first iteration of an updated workflow for our support stewards. The whole idea behind this workflow is to start with the process of standardizing our support response. Historically, our support response has been dependent on the support steward and bias towards a reactive response we do not control properly. This simple updated workflow tries to battle that bias whereas it is bringing a common pattern so all of our support responses are cohesive and shared among our support stewards.

First shared in this Slack message: https://2i2c.slack.com/archives/C028WU9PFBN/p1691789689379759

### DIFF
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -182,12 +182,14 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
 1. `Who: Support steward`
 
    **Respond within 24 working hours**. Acknowledge receipt of the support request and let the {term}`Community Representative` know a time-boxed investigation will start soon. Please request any additional information you may need to be able to reproduce the issue in step 2.
+
 2. `Who: Support steward`
 
    **Spend 30 minutes trying to resolve**. If you believe you can resolve the issue within 30 minutes, try resolving it yourself.
    1. If you resolve the issue, then jump to the "Confirm resolution" step 7.
    2. If you don't believe you can resolve the issue (or you couldn't) in 30 minutes, jump to the next step.
-3. `Who: Support Steward, the board metadata will be revisited by a Partnerships representative and the Engineering Manager`
+
+3. `Who: Support Steward`
 
    **Open an engineering issue**. If this is a {term}`Change Request` or {term}`Guidance Request` and/or you cannot resolve the issue within 30 minutes, then open a support issue for the team to discuss.
 
@@ -200,6 +202,11 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    If the support steward does not have the capacity, they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 6 below).
    
    The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
+
+4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
+
+   ** Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
+     
 5. `Who: Support steward`
 
    **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step 5. Also, move the status of the ticket to the "Pending" state.
@@ -214,20 +221,24 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    this request into our team's work planning process. We'll send an update
    when we've got a plan for completing this request.
    ```
-7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if they are not available`
+
+7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by the Engineering Manager`
 
    **Prioritize the request**. Any non-`critical` issue should wait to be included in our sprints (on Wednesdays, every other week) to be worked out as part of the "planned" reactive work. Follow the [how to prioritize Change and Guidance Requests guide](support:prioritize-requests) to decide how we should prioritize this request relative to the other work we need to do. We should be fully transparent about the support queue to our Community Representatives if they ping us for updates.
    
    If there is any `critical` issue, we could assign people on the fly (during the sprint) to resolve them, but we should minimize that behavior (it should be exceptional cases).
-9. `Who: Support steward`
+
+8. `Who: Support steward`
 
    **Resolve the request**. When some engineer is assigned to a support-related GH issue in the context of a sprint, we move ahead with the investigation/resolution for one (1) sprint. If we failed to find a fix during that time, we communicate back that state in the Freshdesk ticket and resolve it.
 
    Exceptional tickets might need more than one sprint. These tickets need to be explicitly approved as exceptions.
-10. `Who: Support steward`
+
+9. `Who: Support steward`
 
    **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
-11. **Who: Support steward**
+
+11. `Who: Support steward`
 
    **Close the request**. If the Community Representative confirms that their request has been fulfilled, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
 
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -209,7 +209,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
      
 5. `Who: Support steward`
 
-   **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step 5. Also, move the status of the ticket to the "Pending" state.
+   **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step 6. Also, move the status of the ticket to the "Pending" state.
    
 6. `Who: Support steward`
 
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -214,7 +214,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    this request into our team's work planning process. We'll send an update
    when we've got a plan for completing this request.
    ```
-7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if he is not available`
+7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if they are not available`
 
    **Prioritize the request**. Any non-`critical` issue should wait to be included in our sprints (on Wednesdays, every other week) to be worked out as part of the "planned" reactive work. Follow the [how to prioritize Change and Guidance Requests guide](support:prioritize-requests) to decide how we should prioritize this request relative to the other work we need to do. We should be fully transparent about the support queue to our Community Representatives if they ping us for updates.
    
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -197,7 +197,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    
    If the support steward has the capacity, they should self-assign the `critical` issue and work on it (this is now outside of the 30 min timebox in step 2).
    
-   If the support steward does not have the capacity, they should ping the **Engineering Manager** (or the delegated person) so he can secure resources to resolve that issue on the fly (see step 6 below).
+   If the support steward does not have the capacity, they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 6 below).
    
    The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
 5. `Who: Support steward`
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -224,7 +224,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    **Resolve the request**. When some engineer is assigned to a support-related GH issue in the context of a sprint, we move ahead with the investigation/resolution for one (1) sprint. If we failed to find a fix during that time, we communicate back that state in the Freshdesk ticket and resolve it.
 
    Exceptional tickets might need more than one sprint. These tickets need to be explicitly approved as exceptions.
-11. **Who: Support steward**
+11. `Who: Support steward`
 
    **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
 11. **Who: Support steward**
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -197,6 +197,20 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    
    If the issue has a `critical` impact (we defer that first evaluation to the support steward), an additional ping to the support Slack channel is needed to boost the signal.
    
+   ```{admonition} What does `critical` mean?
+
+   We recognize there might be some support-related issues that do not count as [incidents](incidents:what), but 
+   they need a quick resolution (inside the current sprint window) because they are impacting the execution of 
+   desired or existing workflows (degraded experience) for our communities.
+   Examples of those sorts of issues (requests) are:
+     * Image refs updates
+     * Profile updates
+     * User storage limitations
+     * Grafana (and Prometheus) failures
+   Additionally and depending on the nature AND context of the issue (request):
+     * Access to specific buckets
+     * Authentication and authorization updates
+   
    The support steward **should** self-assign the `critical` issue and work on it immediately (this is now outside of the 30-minute timebox described in step 2).
    
    If the support stewards (both of them) do not have the capacity to resolve the `critical` issue (ie. working on another `critical` issue, being out of their working time, etc.), they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 7 below).
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -219,7 +219,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
 
 4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
 
-   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
+   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues. Currently, we allocate a 30-minute working session every Wednesday (open to everyone to participate) to perform such revision.
      
 5. `Who: Support steward`
 
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -148,7 +148,7 @@ Our canned responses.
 We have a few channels for communicating around support requests:
 
 - Our [FreshDesk account](https://2i2c.freshdesk.com/a/) allows for internal team communication via the {guilabel}`Add Note` button. This can be useful for sharing quick internal updates.
-- The [Eng & Prod board that collects support-related issues](https://github.com/orgs/2i2c-org/projects/22/views/47) adding type: `support` metadata to each of our support-related issues. Support requests from external stakeholders are created via emailing `support@2i2c.org`, we do not encourage them to open issues directly.
+- The [Eng & Prod board that collects support-related issues](https://github.com/orgs/2i2c-org/projects/22/views/47) identified via metadata `type==support`.
 - (deprecated) [Issues with the {guilabel}`support` label](https://github.com/2i2c-org/infrastructure/issues?q=is%3Aopen+label%3Asupport+sort%3Aupdated-desc) were where we tracked support requests related to {term}`Change Requests` and {term}`Guidance Requests`.
 
 ## Process for support triage and resolution
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -224,7 +224,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    **Resolve the request**. When some engineer is assigned to a support-related GH issue in the context of a sprint, we move ahead with the investigation/resolution for one (1) sprint. If we failed to find a fix during that time, we communicate back that state in the Freshdesk ticket and resolve it.
 
    Exceptional tickets might need more than one sprint. These tickets need to be explicitly approved as exceptions.
-11. `Who: Support steward`
+10. `Who: Support steward`
 
    **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
 11. **Who: Support steward**
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -205,7 +205,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
 
 4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
 
-   ** Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
+   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
      
 5. `Who: Support steward`
 
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -229,7 +229,7 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
    **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
 11. **Who: Support steward**
 
-   **Close the request**. If the Community Representative confirms that their request has been fulfilled, or if we have not gotten a response after 48 working hours, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
+   **Close the request**. If the Community Representative confirms that their request has been fulfilled, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
 
 (support:prioritize-requests)=
 ## Prioritizing non-incident support requests
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -148,7 +148,7 @@ Our canned responses.
 We have a few channels for communicating around support requests:
 
 - Our [FreshDesk account](https://2i2c.freshdesk.com/a/) allows for internal team communication via the {guilabel}`Add Note` button. This can be useful for sharing quick internal updates.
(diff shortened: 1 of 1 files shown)


sgibson91:
+1 from me

projects/managed-hubs/support.md
```diff
@@ -148,7 +148,7 @@ Our canned responses.
...
 
 - Our [FreshDesk account](https://2i2c.freshdesk.com/a/) allows for internal team communication via the {guilabel}`Add Note` button. This can be useful for sharing quick internal updates.
-- [Issues with the {guilabel}`support` label](https://github.com/2i2c-org/infrastructure/issues?q=is%3Aopen+label%3Asupport+sort%3Aupdated-desc) are where we track support requests related to {term}`Change Requests` and {term}`Guidance Requests`. (support requests from external stakeholders are created via emailing `support@2i2c.org`, we do not encourage them to open issues directly.)
+- [Issues with the {guilabel}`support` label](https://github.com/2i2c-org/infrastructure/issues?q=is%3Aopen+label%3Asupport+sort%3Aupdated-desc) were where we tracked support requests related to {term}`Change Requests` and {term}`Guidance Requests`. Recently, we started to use the [Eng & Prod board to collect support-related issues](https://github.com/orgs/2i2c-org/projects/22/views/47) adding type: `support` metadata to each of our support-related issues. Support requests from external stakeholders are created via emailing `support@2i2c.org`, we do not encourage them to open issues directly.
```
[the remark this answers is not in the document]
damianavila: I like the deprecation idea!

projects/managed-hubs/support.md
```diff
@@ -173,24 +173,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
-4. **Communicate status**. Once we have an issue created to track the next steps, send a message to the Community Representative letting them know about the situation and what the next steps will be. Here's a template to help guide you:
+The goal of the non-incident response process is to bring standardization to our support response. This simple workflow tries to battle the bias towards a reactive response whereas it is also bringing some common patterns so all of our non-incident support responses are cohesive and shared among our support stewards.
+The current iteration of the workflow states each step and who should be responsible/accountable for the specific step, plus some other clarifications.
+
+When a new ticket lands in Freshdesk under the support group and it is not an incident, you should follow the following steps:
+
+1. `Who: Support steward`
+
+   **Respond within 24 working hours**. Acknowledge receipt of the support request and let the {term}`Community Representative` know about any investigation we have done thus far.
```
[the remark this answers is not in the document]
damianavila: I think we need to modify the last piece: "know about any investigation we have done thus far"
The first step should be always ack'ing the ticket, IMHO.
Let me try modifying that last piece of the sentence (I agree it does not make sense as it is written).

projects/managed-hubs/support.md
```diff
@@ -173,24 +173,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+2. `Who: Support steward`
+
+   **Spend 30 minutes trying to resolve**. If you believe you can resolve the issue within 30 minutes, try resolving it yourself.
+   1. If you resolve the issue, then jump to the "Confirm resolution" step 7.
+   2. If you don't believe you can resolve the issue (or you couldn't) in 30 minutes, jump to the next step.
+3. `Who: Support Steward, the board metadata will be revisited by a Partnerships representative and the Engineering Manager`
```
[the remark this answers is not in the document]
damianavila: Good suggestion, let me try something about that to bring more specificity/accountability.

projects/managed-hubs/support.md
```diff
@@ -173,24 +173,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+
+   [{bdg-primary}`Open a general issue`](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=&template=01_new-issue.yml) that describes the request and any next steps we should take. It should be automatically added to the **Eng & Prod** board by the existing automation. Add the **type**: `support` and choose an **impact** level in the **Eng & Prod** board metadata (available for the very same issue, no need to visit the board itself!).
+   
+   If the issue has a `critical` impact (we defer that first evaluation to the support steward), an additional ping to the support Slack channel is needed to boost the signal.
+   
+   If the support steward has the capacity, they should self-assign the `critical` issue and work on it (this is now outside of the 30 min timebox in step 2).
+   
+   If the support steward does not have the capacity, they should ping the **Engineering Manager** (or the delegated person) so he can secure resources to resolve that issue on the fly (see step 6 below).
```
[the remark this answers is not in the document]
damianavila: Mmm, quite interesting... I see your point. I would be towards having false positives to reduce toil, which means advocating to always give the support steward the green light if there is something critical.
Let me think about it a little bit more...

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
-4. **Communicate status**. Once we have an issue created to track the next steps, send a message to the Community Representative letting them know about the situation and what the next steps will be. Here's a template to help guide you:
+The goal of the non-incident response process is to bring standardization to our support response. This simple workflow tries to battle the bias towards a reactive response whereas it is also bringing some common patterns so all of our non-incident support responses are cohesive and shared among our support stewards.
+The current iteration of the workflow states each step and who should be responsible/accountable for the specific step, plus some other clarifications.
+
+When a new ticket lands in Freshdesk under the support group and it is not an incident, you should follow the following steps:
+
+1. `Who: Support steward`
+
+   **Respond within 24 working hours**. Acknowledge receipt of the support request and let the {term}`Community Representative` know about any investigation we have done thus far.
```
damianavila, line 184:
```suggestion
   **Respond within 24 working hours**. Acknowledge receipt of the support request and let the {term}`Community Representative` know a time-boxed investigation will start soon. Please request any additional information you may need to be able to reproduce the issue in step 2.
```

@GeorgianaElena, WDYT?

GeorgianaElena: I like it very much! Thanks @damianavila

projects/managed-hubs/support.md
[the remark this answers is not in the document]
consideRatio: > Having this as a choice still keeps some of the decision toil over the head of support stewards.

:+1: 

---

I favor Georgianas first option of "always prioritizing critical support tasks over any other planned task of a support steward" because I would find the alternative too stressful - leaving a "critical" task hanging while trying to focus on something else work related.

With that said, I'm now realize I'm thinking "critical" is something like an incident. Is "critical" something we use to mean "incident" really? If so, then we should instead rely on such procedure...

Hmmm I guess "critical" could refer to "this has to be done very very soon" and not be an incident describing something going wrong right now.

Tricky topic!

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
    this is not related to an outage or diminished service, so we'll incorporate
    this request into our team's work planning process. We'll send an update
    when we've got a plan for completing this request.
    ```
-5. **Prioritize the request**. Work with our team to decide how we should prioritize this request relative to the other work we need to do. Follow the [how to prioritize Change and Guidance Requests guide](support:prioritize-requests).
-6. **Resolve the request**. Our team should work together to prioritize and resolve this request according to its impact.
-7.  **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
-8.  **Close the request**. If the Community Representative confirms that their request has been fulfilled, or if we have not gotten a response after 48 working hours, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
+7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if he is not available`
```
consideRatio, line 217:
```suggestion
7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if they are not available`
```

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+
+   **Prioritize the request**. Any non-`critical` issue should wait to be included in our sprints (on Wednesdays, every other week) to be worked out as part of the "planned" reactive work. Follow the [how to prioritize Change and Guidance Requests guide](support:prioritize-requests) to decide how we should prioritize this request relative to the other work we need to do. We should be fully transparent about the support queue to our Community Representatives if they ping us for updates.
+   
+   If there is any `critical` issue, we could assign people on the fly (during the sprint) to resolve them, but we should minimize that behavior (it should be exceptional cases).
+9. `Who: Support steward`
+
+   **Resolve the request**. When some engineer is assigned to a support-related GH issue in the context of a sprint, we move ahead with the investigation/resolution for one (1) sprint. If we failed to find a fix during that time, we communicate back that state in the Freshdesk ticket and resolve it.
+
+   Exceptional tickets might need more than one sprint. These tickets need to be explicitly approved as exceptions.
+11. `Who: Support steward`
```
consideRatio, line 227:
```suggestion
10. `Who: Support steward`
```

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+
+   **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
+11. **Who: Support steward**
+
+   **Close the request**. If the Community Representative confirms that their request has been fulfilled, or if we have not gotten a response after 48 working hours, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
```
consideRatio, line 232: If possible we could add automation to transition "resolved" to "closed" after 48 working hours, but I'd say we are fine without it as well. I don't think we should add a manual task to transition resolved to closed.

I've seen both "Resolved" and "Closed" as ticket statuses both meaning that they are finalized, but that "Resolved" indicated that the ticket was finalized by taking successful action or similar, while "Closed" indicated that the ticket was finalized in some other way.

I don't think there is much importance for us to distinguish this, and I don't mind that we have different meanings.

```suggestion
11. **Who: Support steward**

   **Close the request**. If the Community Representative confirms that their request has been fulfilled, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
```

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,62 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
-7.  **Confirm resolution**. Once we have resolved a support request, send a message to the Community Representative to confirm that we believe it is resolved. In FreshDesk, mark the incident as {guilabel}`Resolved`.
-8.  **Close the request**. If the Community Representative confirms that their request has been fulfilled, or if we have not gotten a response after 48 working hours, consider this request closed. In FreshDesk, mark the incident as {guilabel}`Closed`.
+7. `Who: Engineering Manager (currently assigning reactive work) or someone delegated by Engineering Manager if he is not available`
+
+   **Prioritize the request**. Any non-`critical` issue should wait to be included in our sprints (on Wednesdays, every other week) to be worked out as part of the "planned" reactive work. Follow the [how to prioritize Change and Guidance Requests guide](support:prioritize-requests) to decide how we should prioritize this request relative to the other work we need to do. We should be fully transparent about the support queue to our Community Representatives if they ping us for updates.
```
consideRatio, line 219: I don't a clear simple alternative strategy to propose, but this part is the biggest painpoint for me.

Imagine there is a support ticket with that I consider "High" priority that I think I could resolve in 90 minutes for example, then I'd stress out about delegating it to be resolved in the future:
- could cause problem as its "High" priority being considered first in up to two weeks time
- will cause additional overhead work for everyone involved

Those situations make me _very itchy_ to see it resolved directly instead.

Writing this up makes me think that a key issue with the procedure is that we:
- only distinguish non-critical from critical in our response, not low-high prio
- don't have a workflow that relate at all to the duration until the next time we allocate task, even though its probably relevant

I'm imagining a High prio request arriving after a sprint meeting, and the support steward isn't able to resolve it in 30 minutes but perhaps in 90 minutes, then its not comfortable to be a support steward.

consideRatio:
Super happy about seeing this iterated on, I find it very tricky to define a workflow that feels good to follow strictly.

consideRatio (approved):
I'm marking this as approved as I don't want my comments to be seen as a blocker for merging this.

I suggest we just leave un-addressed review comments without marking them as resolved if merge before addressing them, that allows a future iteration to link to them more easily.

projects/managed-hubs/support.md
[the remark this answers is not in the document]
choldgraf: I agree that "acknowledge receipt of the ticket" is a good first step, regardless of the next actions. That way the person knows they are heard, and that folks are thinking about the ticket. This should be a very lightweight / low-cost response.

projects/managed-hubs/support.md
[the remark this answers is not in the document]
choldgraf: My first intuition is that "critical" should trigger the "incident response" process unless it can be resolved really quickly. And follow [our definition of an incident](https://compass.2i2c.org/projects/managed-hubs/incidents/#what-counts-as-an-incident) to define a "critical" support ticket.

That said, I understand that in some cases, we might want a non-incident to be considered critical. For example if a community needs help with something that only affects a small number of people, but it's a really important community to help in that moment as defined by Partnerships.

Either way, we want "critical" to be a fairly high bar - if the bar is too low then everything will be labeled critical and we'll re-create our current pattern of reactiveness.

projects/managed-hubs/support.md
choldgraf: To me, the point is that sprint planning is the moment when we consider task importance + difficulty and turn that into priorities and a plan. If we encourage people to respond to support issues without going through the "zoom out, look at all of our work, now decide what is most important" process, I suspect that the support steward will feel that itchiness and bias themselves towards operational support toil, which is what we're trying to avoid here.

Another reason for this: we want to have mechanisms that encourage the team to share knowledge about operating our infrastructure. If engineers are reacting and resolving immediately, it will be harder to intentionally pair engineers together with the purpose of learning. My hope would be that in our sprint planning, we might say things like "support ticket X is probably going to touch Y part of our infrastructure, so let's pair engineers A and B so that B can see how this is resolved".

choldgraf (commented):
This seems reasonable to try to me - I left a few comments below as well. I think the biggest thing to align everybody on is the idea of "don't just resolve the ticket yourself unless you think it is critical so that we can prioritize it in the next sprint". That feels like the biggest challenge compared with our current mode of doing things.

consideRatio:
Do we equate a freshdesk support ticket's Priority with a opened github issue of `type=support` with the metadata `Impact`?

GitHub | Freshdesk
-|-
image | image

projects/managed-hubs/support.md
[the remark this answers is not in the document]
damianavila: > That said, I understand that in some cases, we might want a non-incident to be considered critical.

This is the main distinction when I am talking about something being "critical", I am not talking about incidents (which we know are critical by definition), I am talking about tasks that need to be done during the current sprint, on the fly, because of X reasons.

> Either way, we want "critical" to be a fairly high bar - if the bar is too low then everything will be labeled critical and we'll re-create our current pattern of reactiveness.

I totally concur, it should be an exception as I mentioned in the text.

projects/managed-hubs/support.md
damianavila: > "zoom out, look at all of our work, now decide what is most important" 
I generally concur with @choldgraf.

Btw, the idea of having "critical" vs. "high-medium-low" tries to make the "critical" state, and hence, the immediate reaction as "high bar" as possible. If there is something "High" in priority that can be solved in 90 minutes, then it should be scheduled in the future (next sprint) OR, if we have **enough time** in the current sprint, and after completing the goal-oriented task, it could be scheduled on the fly as "planned reactive work" (accessing a curated list of "High" issues existing in the current sprint board: https://github.com/orgs/2i2c-org/projects/21/views/4). But again, that should be a low-probability scenario.

Additionally, this conversation also raised the point that we need a better process to put tasks in the planned reactive bucket for specific sprints. Hopefully, we can start planning that using the view I shared above.

projects/managed-hubs/support.md
damianavila: > I don't think there is much importance for us to distinguish this, and I don't mind that we have different meanings.

I tend to agree given the current state of things.

damianavila:
> Do we equate a freshdesk support ticket's Priority with a opened github issue of type=support with the metadata Impact?

I do not think we equate both and that sort of makes sense since we are talking about different territories.
In general, I think there should be 1:1 correspondence/mapping, but I can see times when that differs.
Btw, I also ack that we are using impact in GitHub as a priority scale and we should change that in the future.

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,73 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+   
+   The support steward **should** self-assign the `critical` issue and work on it immediately (this is now outside of the 30-minute timebox described in step 2).
+   
+   If the support stewards (both of them) do not have the capacity to resolve the `critical` issue (ie. working on another `critical` issue, being out of their working time, etc.), they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 7 below).
+   
+   The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
+
+4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
+
+   ** Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
```
damianavila, line 208:
```suggestion
   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
```

projects/managed-hubs/support.md
```diff
@@ -181,25 +181,32 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
...
    2. If you don't believe you can resolve the issue (or you couldn't) in 30 minutes, jump to the next step.
-3. `Who: Support Steward, the board metadata will be revisited by a Partnerships representative and the Engineering Manager`
+
+3. `Who: Support Steward`
 
    **Open an engineering issue**. If this is a {term}`Change Request` or {term}`Guidance Request` and/or you cannot resolve the issue within 30 minutes, then open a support issue for the team to discuss.
 
    [{bdg-primary}`Open a general issue`](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=&template=01_new-issue.yml) that describes the request and any next steps we should take. It should be automatically added to the **Eng & Prod** board by the existing automation. Add the **type**: `support` and choose an **impact** level in the **Eng & Prod** board metadata (available for the very same issue, no need to visit the board itself!).
    
    If the issue has a `critical` impact (we defer that first evaluation to the support steward), an additional ping to the support Slack channel is needed to boost the signal.
```
choldgraf, line 198: Can you cross-link to a definition of "critical"? I think without this people will have different ideas of what "critical" means. It is probably similar to "what is an incident" but we should make that explicit and clarify when something can be critical even if not an incident

choldgraf (approved):
These improvements look good to me. I think if there are no objections we should merge this and start following it, under the assumption that the Partnerships team will likely need to make modifications when they take over stewardship of this process.

yuvipanda:
There are two primary questions I have:

1. Does this mean our SLO for responding to 'non critical support tickets' is upto 2 weeks?
2. There's no more sprint planning every other week, as sprints are organized per-week. So *when* does this prioritization actually happen? I'd like that detail to be fixed before this gets merged.

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,73 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+   
+   The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
+
+4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
+
+   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
+     
+5. `Who: Support steward`
+
+   **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step 5. Also, move the status of the ticket to the "Pending" state.
```
sgibson91, line 212:
```suggestion
   **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step ???. Also, move the status of the ticket to the "Pending" state.
```

This currently is step 5, so "in step 5" here doesn't make sense

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,73 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+   1. If you resolve the issue, then jump to the "Confirm resolution" step 7.
+   2. If you don't believe you can resolve the issue (or you couldn't) in 30 minutes, jump to the next step.
+
+3. `Who: Support Steward`
+
+   **Open an engineering issue**. If this is a {term}`Change Request` or {term}`Guidance Request` and/or you cannot resolve the issue within 30 minutes, then open a support issue for the team to discuss.
+
+   [{bdg-primary}`Open a general issue`](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=&template=01_new-issue.yml) that describes the request and any next steps we should take. It should be automatically added to the **Eng & Prod** board by the existing automation. Add the **type**: `support` and choose an **impact** level in the **Eng & Prod** board metadata (available for the very same issue, no need to visit the board itself!).
+   
+   If the issue has a `critical` impact (we defer that first evaluation to the support steward), an additional ping to the support Slack channel is needed to boost the signal.
```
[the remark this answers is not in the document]
damianavila: @2i2c-org/engineering, what do people think about the above list for "critical" support issues?
Am I missing something? Am I adding too many items? Thoughts?

projects/managed-hubs/support.md
damianavila:
```suggestion
   **Add a reference/link to the created engineering issue inside the Freshdesk ticket**. You can use an internal note or make it public when you communicate back to the Community Representative in step 6. Also, move the status of the ticket to the "Pending" state.
```

projects/managed-hubs/support.md
damianavila: Tried it at https://github.com/2i2c-org/team-compass/pull/764#discussion_r1315099927.
Waiting for feedback from folks about it.

damianavila:
Thanks for your feedback, @yuvipanda!

>Does this mean our SLO for responding to 'non critical support tickets' is upto 2 weeks?

Indeed, that is a realistic SLO for non-critical support tickets, IMHO.
Critical ones would have an SLO below 2 weeks because it can be assigned on the fly in the current sprint.

>There's no more sprint planning every other week, as sprints are organized per-week. So when does this prioritization actually happen? I'd like that detail to be fixed before this gets merged.

I am usually doing a little bit of prioritization on Wednesdays when a new sprint starts.
In this proposal, we are saying we are doing prioritization every week, so I will create a 30-minute spot every Wednesday as a sort of working session for the ones interested in joining me in the prioritization task. It would be great to have at least one participant from @2i2c-org/partnerships-and-community-guidance because they will surely help us to properly prioritize according to community needs/impact.

damianavila:
FYI (and given the feedback already provided) I will merge this PR by Tuesday 5th EOD unless there is an explicit blocker.
Thanks!

jmunroe:
I will join a Wednesday prioritization session.  That's a good idea.

projects/managed-hubs/support.md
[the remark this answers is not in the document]
consideRatio: Not sure if I think access to specific buckets should be considered critical, and I could see various authentication and authorization updates as critical and non-critical depending on their nature.

"Image updates" are a bit vague in my mind since I'm thinking we are not responsible for the user environment images - unless we are in certain exceptions. Is our agreements in those exceptions that we should help quickly as well?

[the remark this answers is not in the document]
damianavila: I think access to buckets, auth and authz can be considered critical depending on their nature. So, I should probably clarify that further.
With "image updates" I am only trying to refer to the update on the reference to an image, such as https://github.com/2i2c-org/infrastructure/pull/3034/files, I should probably be more specific about it!
Thanks for the feedback!! I will update the list accordingly.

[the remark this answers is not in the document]
damianavila: @consideRatio, wdyt about this one?
```
Examples of those sorts of issues (requests) are:
     * Image refs updates
     * Profile updates
     * User storage limitations
     * Grafana (and Prometheus) failures
Additionally and depending on the nature AND context of the issue (request):
     * Access to specific buckets
     * Authentication and authorization updates
```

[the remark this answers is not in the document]
consideRatio: Looks great @damianavila!!

yuvipanda:
> In this proposal, we are saying we are doing prioritization every week, so I will create a 30-minute spot every Wednesday as a sort of working session for the ones interested in joining me in the prioritization task.

This sounds good to me @damianavila - I am not sure if this is written into the proposal, but if not it should be and then we can try that out.

projects/managed-hubs/support.md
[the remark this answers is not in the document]
damianavila: Updating the suggestion as per the previous message and committing now.

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,87 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+   
+   The support steward **should** self-assign the `critical` issue and work on it immediately (this is now outside of the 30-minute timebox described in step 2).
+   
+   If the support stewards (both of them) do not have the capacity to resolve the `critical` issue (ie. working on another `critical` issue, being out of their working time, etc.), they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 7 below).
+   
+   The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
+
+4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
+
+   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues.
```
damianavila, line 222:
```suggestion
   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues. Currently, we allocate a 30-minute working session every Wednesday (open to everyone to participate) to perform such prioritization.
```

damianavila: More specificity on the revision working session as per suggestion from @yuvipanda.

damianavila:
> This sounds good to me @damianavila - I am not sure if this is written into the proposal, but if not it should be and then we can try that out.

It was conceptually there. Even when I see the specific of the format as an implementation detail, I have added a [line](https://github.com/2i2c-org/team-compass/pull/764/commits/f5bb850cb974ca4fb464a93473b3b8b756cb2954) explaining the proposed format as I described above (that we can change in the future if we realize this is not working as intended).

projects/managed-hubs/support.md
```diff
@@ -173,24 +174,89 @@ This process is carried out in an ongoing basis by the {term}`Support Stewards`.
...
+   
+   The support steward **should** self-assign the `critical` issue and work on it immediately (this is now outside of the 30-minute timebox described in step 2).
+   
+   If the support stewards (both of them) do not have the capacity to resolve the `critical` issue (ie. working on another `critical` issue, being out of their working time, etc.), they should ping the **Engineering Manager** (or the delegated person) so they can secure resources to resolve that issue on the fly (see step 7 below).
+   
+   The support steward **should not** work on issues with impact lower than `critical` (unless they are assigned as part of the "planned" reactive work in the context of a running sprint (see step 6 below).
+
+4. `Who: Partnerships representative and the Engineering Manager (or respective delegates)`
+
+   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues. Currently, we allocate a 30-minute working session every Wednesday (open to everyone to participate) to perform such revision.
```
damianavila, line 224:
```suggestion
   **Revisit the impact metadata**. Once a week (at minimum) the [support view in the **Eng & Prod** board](https://github.com/orgs/2i2c-org/projects/22/views/47) should be revisited to validate the impact level on support-related issues. Currently, we allocate a 30-minute working session every Wednesday (open to everyone to participate) to perform such impact revision and further prioritization ("planned" reactive) every other week (see step 7 for more details).
```

PR #772, opened 2023-10-21
Fix broken URLs

AIDEA775:
- Fixed and migrated all old URLs starting with "/en/latest/".
- Update new-team-member template


choldgraf (approved):
Many thanks! I added a couple extra fixes for broken links following the same pattern. LGTM

PR #779, opened 2023-11-14
Update response-protocol.md

haroldcampbell:
Fixed typo so that the clause for dealing with violations outside of the standard process is clear.


choldgraf (approved):
Thanks for this! It reads much clearer now

PR #781, opened 2023-11-22
Add 30m support ticket evaluation checklist

GeorgianaElena:
Companion of https://github.com/2i2c-org/infrastructure/pull/3366.

At the moment it is more of a placeholder page with links to the engineering troubleshooting section. But it is a page where a checklist of things to do in the first 30m a ticket is checked can live.

When separating first line from second line support, I imagine this page should be a guidebook that can be followed by the first line support before escalating the ticket.


sgibson91 (approved):
❤️

PR #791, opened 2024-01-08
Update non incident docs to match new workflow

GeorgianaElena:
Once https://github.com/2i2c-org/infrastructure/pull/3586 gets merged, these tasks will happen automatically. This is a follow-up PR to update the docs.

----
📚 Documentation preview 📚: https://2i2c-team-compass--791.org.readthedocs.build/en/791/

### DIFF
--- a/projects/managed-hubs/support.md
+++ b/projects/managed-hubs/support.md
@@ -193,10 +193,12 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
 
 3. `Who: Support Steward`
 
-   **Open an engineering issue**. If this is a {term}`Change Request` or {term}`Guidance Request` and/or you cannot resolve the issue within 30 minutes, then open a support issue for the team to discuss.
+   **Open an engineering issue**. If this is an issue that cannot be resolved within 30 minutes, then open a support issue for the team to discuss.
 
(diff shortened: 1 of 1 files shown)


projects/managed-hubs/support.md
```diff
@@ -193,10 +193,12 @@ When a new ticket lands in Freshdesk under the support group and it is not an in
 
 3. `Who: Support Steward`
 
-   **Open an engineering issue**. If this is a {term}`Change Request` or {term}`Guidance Request` and/or you cannot resolve the issue within 30 minutes, then open a support issue for the team to discuss.
+   **Open an engineering issue**. If this is an issue that cannot be resolved within 30 minutes, then open a support issue for the team to discuss.
+
+   [{bdg-primary}`Open a Freshdesk ticket tracker issue`](https://github.com/2i2c-org/infrastructure/issues/new?assignees=&labels=&template=5_freshdesk-ticket.yml) that describes the ticket and provide as many details as possible about the results obtained in the first 30m investigation (if any).
```
sgibson91, line 198: I think these sentences could be made a little clearer. We have "engineering issue", "support issue" and "Freshdesk ticket tracker issue" - are they all the same, different?

I also have a small quibble about using "Freshdesk" in our naming when in the future we could decide to move away from Freshdesk. But let's not block on that particularly.

sgibson91 (approved):

projects/managed-hubs/support.md
GeorgianaElena: Good point @sgibson91! Thank you! Do you think https://github.com/2i2c-org/team-compass/pull/791/commits/22849641f67de77d7c98a3c25adf4921fa618757 made it better?

sgibson91: Yeah, definitely! It reads like we're only talking about one issue now instead of three!

PR #794, opened 2024-01-18
Add DM and PM role descriptions

choldgraf:
This PR adds an interrim job description for the Delivery Manager / Chief of Staff, as well as the Product Lead roles. It does so by re-using the text from the job postings for each role. I simply copy/pasted and converted each to markdown.

We will need to define more formal definitions of these roles within 2i2c, including metrics for success and advancement, but for now I wanted to use these job descriptions as a marker / starting point.

----
📚 Documentation preview 📚: https://2i2c-team-compass--794.org.readthedocs.build/en/794/


choldgraf:
Merging this one since we'll need to keep building on some of these changes in the coming weeks, but I want to get the basic content for roles documented now that people are joining in them!

PR #803, opened 2024-02-06
Update explanations of accounting and contracts airtable

choldgraf:
This adds more information about our accounting and contracts airtables, with the goal of making it clearer what those numbers mean.

@colliand can you provide feedback on what changes here could make this information easier to digest and follow? I have a feeling it is too complex and messy but not sure how to improve it.


choldgraf:
I'm going to merge this in since it's an iterative improvement and since Jim is out of town

Issue #820, opened 2024-03-14
Feedback on draft proposal for "Upstream" work


aprilmj:
This is a proposal for how we might make upstream leadership more visible and more shared. This proposal is a bit more experimental than the other board and process proposals (#817 and #819) and attempts to address a range of different types of work. 

### Questions for feedback
1. What's missing or glaringly wrong about what I've captured? 
2. What other alternatives, if any, should we consider? 
3. Is this safe to try, and if so, how do we know it's succeeded or failed? 
4. Please either make changes to this top level comment to change wording, but add new comments with bigger questions - so we have a history of the discussion. 

### Why is this important? 
We want to act as a more effective team, both in developing 2i2c's product and service and in contributing to the wider world of knowledge community infrastructure (trying out the language from the value proposition). If we can see all the work happening in a given period, we'll be clearer about our true capacity for work, make better decisions about how we use our capabilities, and be able to contribute and grow more intentionally.

**Which kinds of work need a "something else" process or board?** 
When we attempted to refine the collection of boards currently used to track work, several similar points came up as needing "something else" - meaning they were not visible in our current boards and processes (from the March Initiative Board refinement meeting):
- Catalyst Board is something else  
- Board for 2i2c to provide leadership upstream?  ???
- JupyterHealth Board  
- JupyterGIS
- We need a tool for cross-functional coordination at the mid-level (ie. between a subset of areas)  
- VEDA board  
- HHMI
- Openscapes
- CryoCloud
- Pythia
- NASA ScienceCore

> @colliand added some more items above. 

The common elements across all these points are - they involve external teams (and may have their own project boards where work is tracked), they result in tasks that need to be completed by Engineering and Partnerships team members, and they often are not visible to the entire 2i2c organization. 

> @colliand adding. These external engagements share: 2i2c providing thought/science/technical/collaboration leadership; 2i2c serving as "glue" to bring otherwise separated groups together; 2i2c not controlling the agenda of the collaboration; 2i2c exposure to scope creep without cost recovery; 

### The proposal: external and upstream leadership initiatives
We already have a source of truth (as we improve it) for non-reactive work: the Initiatives Board. We also have a process proposed for planning Engineering or Operations (including Partnerships) tasks within a sprint. And we have project boards we share with external collaborators. Creating more boards might make this worse; instead, let's add more process to the boards we have. 

**Track every significant upstream/project leadership effort as an initiative**
[...]

PR #830, opened 2024-03-26
Update workflow.md for "community projects"

aprilmj:
Added a bit about community projects (formerly called "upstream", language improved by Erik). 

Please check my description and edit as needed - I'm not positive that I completely understood the different types of community work, and don't want to use misleading language. 

This should close issue #820

### DIFF
--- a/cross-functional/workflow.md
+++ b/cross-functional/workflow.md
@@ -65,6 +65,11 @@ what kind of work it will entail, and why it is important to 2i2c's overall goal
 ```
 ````
 `````
+:::{admonition} **What about community projects?**
+
+We also track community projects as initiatives. When 2i2c is actively engaged in work for upstream projects or communities we serve, we add those to our initiatives list, and follow the planning process outlined here. 
+
+Note this is a work in progress, and we’re continuing to evolve our process for collaborating within and outside 2i2c.  :::
 
 ### Bi-weekly initiatives meetings
 
--- a/cross-functional/workflow.md
+++ b/cross-functional/workflow.md
@@ -65,11 +65,13 @@ what kind of work it will entail, and why it is important to 2i2c's overall goal
 ```
 ````
 `````
-:::{admonition} **What about community projects?**
+:::{admonition} What about community projects?
 
-We also track community projects as initiatives. When 2i2c is actively engaged in work for upstream projects or communities we serve, we add those to our initiatives list, and follow the planning process outlined here. 
+We also track community projects as initiatives.
+When 2i2c is actively engaged in work for upstream projects or communities we serve, we add those to our initiatives list, and follow the planning process outlined here. 
 
-Note this is a work in progress, and we’re continuing to evolve our process for collaborating within and outside 2i2c.  :::
+Note this is a work in progress, and we’re continuing to evolve our process for collaborating within and outside 2i2c.
+:::
 
 ### Bi-weekly initiatives meetings
 


cross-functional/workflow.md
```diff
@@ -65,6 +65,11 @@ what kind of work it will entail, and why it is important to 2i2c's overall goal
 ```
 ````
 `````
+:::{admonition} **What about community projects?**
+
+We also track community projects as initiatives. When 2i2c is actively engaged in work for upstream projects or communities we serve, we add those to our initiatives list, and follow the planning process outlined here. 
+
+Note this is a work in progress, and we’re continuing to evolve our process for collaborating within and outside 2i2c.  :::
```
choldgraf, line 72:
```suggestion
:::{admonition} **What about community projects?**

We also track community projects as initiatives.
When 2i2c is actively engaged in work for upstream projects or communities we serve, we add those to our initiatives list, and follow the planning process outlined here. 

Note this is a work in progress, and we’re continuing to evolve our process for collaborating within and outside 2i2c.
:::
```

choldgraf: just a formatting suggestion

choldgraf (commented):
Thanks for this - a quick thought on the details here:

My understanding is that the process for community initiatives is a bit different from our other initiatives, in that community initiatives define their task list in a separate issue or project board that we cross-reference from our "community initiative" issue. Do others agree that is missing from this description?

choldgraf:
Something else I'm noticing as I look at the initiatives board, where we've now got several community projects:

1. It's not clear how those projects relate to the columns ("needs refinement" etc). Should we just assume that they perpetually stay in the "needs refinement" column, with the idea that each one will generate more specific/actionable initiatives that we then track independently on the board?
2. Many of those projects have dates embedded in their titles, which makes it hard to parse them. Could we instead add `start date` and `end date` fields to these boards for any initiatives that have a timeline we need to keep in mind? This could be something specific like the "period of performance" for contracts, or could be something team-driven like a time box for when we want to conclude an initiative.

cross-functional/workflow.md
aprilmj: I like the change, thanks!

aprilmj:
>     1. It's not clear how those projects relate to the columns ("needs refinement" etc). Should we just assume that they perpetually stay in the "needs refinement" column, with the idea that each one will generate more specific/actionable initiatives that we then track independently on the board?

That seems like a good place to start, based on what @jmunroe described in his Pythia thought experiment. We may realize we don't want to keep the long list in "needs refinement" once we're sure we have them all covered.

>     2. Many of those projects have metadata embedded in their titles (specifically "PROJECT" and dates), which makes it hard to parse them. Could we instead add `start date` and `end date` fields to these boards, and use a "community initiative" field to categorize them? The dates could be something specific like the "period of performance" for contracts, or could be something team-driven like a time box for when we want to conclude an initiative.

+1, that sounds like a great idea. I like that we're figuring out details as we go.

choldgraf:
Are there any objections to using board metadata for that information? In particular if @jmunroe or @colliand may have thoughts since (I think) they are the ones that chose that format for the `PROJECT` issues.

jmunroe:
I think shorter issue names and the use of labels/tags to store start and end dates makes sense.  I've made those changes to the names.

jmunroe:
I merged in Chris reformatting suggestion.

jmunroe (approved):

jmunroe:
I think there is more to discuss about  'community projects' and how to integrate them into our on going work in defining initiatives. But there is nothing blocking merging in this PR since it already makes it clear that this this a work in progress.

@aprilmj I think you can click the 'Merge pull request' when you are ready to do so.

aprilmj:
Thanks everyone! For future reference, what's this team's norm for documentation-related merges? 

I was just thinking that I would be 100% ok with @jmunroe pushing the button while you were looking 👍

choldgraf:
We don't have a formal policy for merging documentation changes, but I think that [following the team's decision making principles around bias towards action](https://compass.2i2c.org/operations/governance/#bias-towards-action), I'd be comfortable encouraging people to "merge first if they think it's safe to try" unless there's something controversial or potentially wrong in there...

PR #832, opened 2024-03-29
Fixed typo in the docs.2i2c.org url

haroldcampbell:


choldgraf (approved):
Thanks! @GeorgianaElena you should feel free to just merge when it's an obvious typo like this! (same to you @haroldcampbell , I think if a change is minor and obviously a fix, let's just merge it in...reviews are useful if there's value in sharing team knowledge about a change, but in this case it is very minor and obviously a typo)

PR #836, opened 2024-04-15
Create Prioritization.md

Gman0909:
Adding a link to the Productboard primer

### DIFF
--- a/product/Prioritization.md
+++ b/product/Prioritization.md
@@ -1,0 +1,5 @@
+# Using Productboard for prioritization
+
+As of February 2024, 2i2c adopted Productboard as a tool to help manage product prioritization and collect insights relevant to our product improvement process. 
+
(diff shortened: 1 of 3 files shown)


choldgraf (approved):
This looks good to me! I made a few minor changes for wording which I'll list below. Also confirmed that this is showing up in our docs build preview. I'll list a few thoughts below just to do some information sharing around our docs workflow in general.

### Docs preview

First, FYI you can preview what documentation changes look like by clicking "show all checks" and then clicking the `Details` link in the "ReadTheDocs" job. Here's a snapshot of the UI flow for this. 

CleanShot 2024-04-15 at 08 52 05@2x

### Markdown link wording

I expanded your markdown link text a bit so that it's easier to identify and understand. Just a minor change but I think it's best practice to make hyperlinks self-descriptive ([here's one discussion on hyperlink text](https://ux.stackexchange.com/questions/28146/how-much-of-the-related-text-should-be-included-in-a-hyperlink) in case it's of interest)

### Added the index.md changes in the same PR

I see you've got another PR open in #837 . I've made that change here as well so that we can have it all in one place. The easiest way to do this with a web-based workflow is to use GitHub's native `vscode` editor. [Here's the `github.dev` repository](https://github.com/github/dev) - it has some instructions for how to do this. Just sharing in case this makes it easier in the future!

PR #837, opened 2024-04-15
Update index.md

Gman0909:
Updating the TOC to include the newly created page linking to the productboard primer


choldgraf:
closing this one as we have implemented it via @Gman0909 's other PR here: #836

choldgraf:
I think (?) this was an accidental re-open, since the content is already live on our site. Let me know if that's not right!

PR #840, opened 2024-04-17
Update overview.md

haroldcampbell:
The scope and responsibilities was outdated.


Gman0909 (approved):

choldgraf:
Thanks @haroldcampbell - I've made one change to the commit so that we cross-link the Product Lead role, and I've also updated the PR title so that it's more descriptive of the change we're making. Thanks!

PR #845, opened 2024-04-18
Describe definition of ready for 3 kinds of tasks

yuvipanda:
Ref https://github.com/2i2c-org/meta/issues/953

### DIFF
--- a/engineering/definition-of-ready.md
+++ b/engineering/definition-of-ready.md
@@ -22,7 +22,7 @@ The following non-exhaustive initial list of the different kinds of tasks that t
 5. [](ready:fault-fixing)
 6. [](ready:community-review)
 7. [](ready:incident-Response)
-8. [](ready:routing)
+8. [](ready:fault-fixing)
 
 (dor:ready)=
 ## Definition of ready
--- a/engineering/definition-of-ready.md
+++ b/engineering/definition-of-ready.md
@@ -65,7 +65,9 @@ be very similar.
 - [ ] Any additional information needed for this specific configuration change.
 
 Functionally, this means that if we are requested to change config for something that we can't
-find any documentation for, we should
+find any documentation for, it should be escalated (via a message to the `#engineering` slack channel)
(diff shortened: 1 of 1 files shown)


engineering/definition-of-ready.md
```diff
@@ -51,16 +51,37 @@ should not be considered a feature we currently support!
...
+
+- [ ] Link to documentation about this feature, which has at least some mention of the possible
+      ways to make configuration changes. Since 2i2c documentation may not cover all the possible
+      options (as that would be too unweildy), this could be either a link to 2i2c documentation on
+      [infrastructure.2i2c.org](https://infrastructure.2i2c.org) or a link to upstream documentation
+      about various configuration options for this feature.
+- [ ] Any additional information needed for this specific configuration change.
+
+Functionally, this means that if we are requested to change config for something that we can't
+find any documentation for, we should
```
consideRatio, line 68: Unfinished sentence

GeorgianaElena (approved):
LGTM

engineering/definition-of-ready.md
yuvipanda: Whoops. I completed this sentence, lmk what you think

consideRatio:
This is tricky to be specific about! Not sure about an action point, but reflected a bit.

1. A config change not documented yet could blur the line of new feature.
2. A config change could be simple and easy, but perhaps involve unsustainable toil, unfair in a shared cluster with regards to sharing costs, or have some reason making it worth deciding on if we should or shouldn't do something like this for communities.

I think its good that this is escalated to #engineering to be decided about by more than one person.

Documenting likely makes this work take more than 30 minutes, so what should the triager do?
- schedule to be done in s sprint
- config change now+document now
- config change now + create an issue to document
- document now + create an issue to config change

Maybe to be decided in engineering slack chat as well?

consideRatio (approved):

sgibson91 (approved):

PR #849, opened 2024-04-18
docs: fail on build warnings, fix warnings, refresh conf.py

consideRatio:
I saw we had many sphinx build warnings that I figured we shouldn't have, because they are typically very relevant and introduced as part of changes into docs. This workflow will ensure we don't introduce warnings.

Sibling PR to https://github.com/2i2c-org/infrastructure/pull/3948


consideRatio:
I'd like these checks to be in place for current open PRs and one I'm about to work on now myself - I think its safe to merge - going for it.

PR #865, opened 2024-06-01
Add guide to having open, honest conversations

choldgraf:
Harold shared a helpful guide on having open and honest conversations in https://github.com/2i2c-org/meta/issues/1159. I thought it might be helpful if we wrote this up in our compass for future reference and discovery.

- closes https://github.com/2i2c-org/meta/issues/1159


choldgraf:
I'm merging this in because it's useful content and we don't want to lose it!

PR #868, opened 2024-06-18
Add fundraising strategy document

choldgraf:
This converts our fundraising strategy document from Google Docs to the team compass so that it is more discoverable. I cleaned up some of the language and tried to add extra information, and added a little diagram as well.

I also copied over the language that @colliand used in https://github.com/2i2c-org/meta/issues/1189 since that felt like a good and already-formed bit of information to use.

Complementary post:

- https://github.com/2i2c-org/2i2c-org.github.io/pull/255


colliand (commented):
Thanks @choldgraf!

colliand:
Hmm. The deployment preview is red and failing.

sgibson91:
> Hmm. The deployment preview is red and failing.

I think this is due to some scripts/notebooks being executed on build that require secrets, and by default secrets aren't available in PRs for security reasons. We should document this and/or provide a way to skip the parts of the build that require secrets in the PR context so we don't become nonchalant about dismissing failing builds and miss an _actual_ failure.

jnywong:
I've fixed the issue in https://github.com/2i2c-org/team-compass/pull/868/commits/80ce215c2ebf2ecc6b28bc5e6254961c8e9bb888.

The page on Grafana monitoring wasn't working as expected as there were a bunch of new 503 errors from hubs that weren't returning any Prometheus queries.

Secrets are unavailable in a _forked_ PR I believe, see the [GH Docs](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow)

jnywong:
@sgibson91 I also created this reference guide for GH secrets https://team-compass.2i2c.org/reference/documentation/secrets/

jnywong:
I am biasing towards action and merging this in given that I presume @colliand has already cast eyes over it above 👀

sgibson91:
@jnywong That's good, but I still think there needs to be a distinction between "This workflow failed because it couldn't access the secret" and "This workflow failed because the change actually broke something". Otherwise we stop trusting our tests, and if we don't trust tests then there's no point in having them.

PR #870, opened 2024-07-08
Add note of screen.studio key

choldgraf:
Documents the screen.studio keys we bought last week


jnywong:
Haha I was going to at the end of my day! You're just too darn quick

PR #890, opened 2024-08-01
Document monthly budget projections process

choldgraf:
This is light documentation about the monthly process that we follow for creating budget projections when CS&S sends us new accounting reports each month.


choldgraf:
Am going to merge this in since it's just pointing to a process we're already following

PR #962, opened 2025-04-25
deprecate old partnerships content

colliand:
The top level folder `biz-dev` was renamed `business-development`. 

The top level index.md file pointed to `biz-dev/index`. This link was changed to point at `business-development/index`. 

Some files from `partnerships` were copied into `business-development`. The remaining out-of-date files were left in `partnerships`. 

The `partnerships` folder was deleted. 

---

Broken link targets found in deployment previews. Further edits to fix those links.


choldgraf:
I've fixed the problems. Here's what I did:

- You deleted some sections but didn't delete references _to_ those sections. Specifically, you'd deleted the "slack" documentation for BD, and to a Freshdesk Footer section (I think because in your mind FreshDesk is no longer a BD responsibility, is that right?)
- The "communication" section was nested under partnerships, but is _not_ nested underneath `business-development`. I've moved it to a top-level section. **is that intentional for you @colliand ?**

I think we're good to merge if the above two points make sense to @colliand

colliand:
My deletions were more chainsaw than scalpel. I'd like to retain links between BD and Freshdesk sections. The way support overlaps with engagement, retention, renewals and referral lead generation is under consideration by me.

colliand:
I welcome your top level restorations. This documentation will evolve over the upcoming sprints.

choldgraf:
Before merging can you confirm that there is nothing else deleted that you don't want to delete?

Keep in mind: the compass is also a reflection of responsibility. If freshdesk is documented within BD, the implication is that BD is accountable and responsible for anything documented therein. If there are processes that others outside of BD are responsible for, but where BD is _accountable_ for, this needs to be clear in the docs.

colliand:
I call out a tension I am feeling here. There are ongoing team-wide discussions about how "customer success" and "engagement" is in a liminal space following the Partnerships/Engineering to BD/DE/P&S reorganization. On one hand, I want BD to not be accountable or responsible for Freshdesk right now due to capacity. On the other hand, I want BD to embrace Freshdesk and the associated task streams as _vital_ to our customer's success and a driver toward renewal.

I will review as best as I am able today and merge.

PR #964, opened 2025-05-02
Document new GL codes

choldgraf:
This documents some custom GL codes that we've started using with CS&S.

closes https://github.com/2i2c-org/meta/issues/2164


choldgraf:
I got confirmation that this looks good!

PR #972, opened 2025-06-04
Update new-team-member.md issue template

GeorgianaElena:
Updates clusters that need access granted

Ref: https://github.com/2i2c-org/meta/issues/2249


jmunroe:
We should also add Jetstream2:

Jetstream Allocation: 
- SEE240014
- CIS250031

PR #973, opened 2025-06-04
Add what what happens when an agreement it signed.

jmunroe:


jmunroe:
I am withdrawing this PR at this time.

PR #974, opened 2025-06-05
Add source of truth information for BD

haroldcampbell:
I'm trying to clarify the different places we store information and how we use: 
- BD Github project board
- Hubspot (Deals/CRM)
- Airtable (Engagement management)


aprilmj (changes requested):
There's one small spelling mistake, and some words capitalized that shouldn't be. Otherwise great, makes sense to me!

PR #976, opened 2025-06-06
Update sales-operations.md to include fine print text requested by P&S

colliand:
BD added fine print text to the 2i2c/CS&S service agreement template to set terms around 2i2c team access to cloud infrastructure. Sometimes in the past, 2i2c's customers asked for PII about our engineers that went beyond our comfort zone. In other cases, customers gated access to infrastructure by requiring our team to use customized accounts managed by the customer. BD will default include this text to help our team avoid these complications in the future.


business-development/sales-operations.md
```diff
@@ -69,8 +69,8 @@ Expanded sales process
 
 ### Agreements
```
choldgraf, line 70:
```suggestion
### Service Agreements template and guidelines
```

choldgraf:
A question: Do you intend to update this section every time you add new text to the template? (ie, is that why you included language like `text was added`?

If not, I suggest thinking about the most important things that need to be mentioned explicitly in the compass, and use a "changelog" in the template docs for recording the important changes you've made there, and why.

jnywong:
@colliand just seen this PR and sending you a ping

colliand (commented):
Thanks. This looks good.

PR #988, opened 2025-06-23
Update high-level sections and workflows to point to sources of truth

choldgraf:
This PR cleans up our team compass so that our high-level organizational information is easier to discover, and to update sources of truth.

- Defines a separate value proposition section so it's more discoverable, and adds a few links to documents that are driving this right now.
- Adds references to [our strategy re-focus doc](https://docs.google.com/document/d/1HgSqAWeZZjcph3eHuWfv56WYopEHwrtmb-pAL3wMxco/edit?tab=t.0) wherever possible. We can update the compass when the text has stabilized.
- Adds some process information from the last quarterly planning process.
- Documents the accountable/responsible party team practice we've been following lately


choldgraf:
This is largely about updating the compass to reflect the current state of our team conversations, it's not proposing any new information. For this reason, I suggest we merge this one in without review being necessary, and focus on the strategy-realigning document to get on the same page about our overall goals etc.

PR #994, opened 2025-07-16
Update documentation for BD

haroldcampbell:
- Added brief note about how identify which Engagements need renewing
- Restructured BD document hierarchy so that it's easier to find information

### DIFF
--- a/communication/talks.md
+++ b/communication/talks.md
@@ -10,14 +10,17 @@ If you publish your talk on Zenodo, please add it to [our Zenodo community](publ
 
 ## When to give talks
 
-If a talk is virtual and a "standard" talk length, feel free to accept these opportunities to signal-boost our mission and impact.
-If a talk is in-person and requires travel that is not reimbursed by the organizers, or if it is particularly complex or important, get approval from a team lead at 2i2c before giving the talk.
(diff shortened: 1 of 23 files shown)


choldgraf:
I'm going to merge this one because it seems reasonable and more like improving/rearranging rather than adding new policy. I suggest that y'all continue to improve these docs iteratively and just self-merge this kind of thing in the future.

choldgraf:
Actually I take that back - some decisions in this PR don't make sense to me and they've broken the docs build.

@haroldcampbell why did you move a bunch of pages under a `business-development/bd/` folder? What's the benefit of `bd/`?

CleanShot 2025-08-29 at 10 55 51@2x

If we do this, we'll need to update all relative paths in all of those pages so that they aren't broken. That's causing the docs build to fail.

choldgraf (changes requested):
Actually I take that back - some decisions in this PR don't make sense to me and they've broken the docs build.

@haroldcampbell why did you move a bunch of pages under a `business-development/bd/` folder? What's the benefit of `bd/`?

CleanShot 2025-08-29 at 10 55 51@2x

If we do this, we'll need to update all relative paths in all of those pages so that they aren't broken. That's causing the docs build to fail.

PR #996, opened 2025-07-23
Update engineering structure 

aprilmj:
Adds an important note that we treat the Engineering Manager and Technology Lead as co-equals in hiring and exit decisions, since we each have a part of the information needed to hire or exit an engineer. Practically that means we must either reach a consensus or disagree but commit to the other person's decision.


choldgraf (approved):
Thanks for the clarification!

yuvipanda (approved):

PR #1000, opened 2025-07-29
Explicitly prioritize incident action items

yuvipanda:
One of our antipatterns is to not take action on our incident action items, because they do not explicitly take priority. So they get left for 'some day' in the future, and unfortunately often that means they cause another outage. This was true for the cloudbank outage - this had happened before, but it recovered itself, so we left it untouched.

We explicitly add process here to make sure we bring this work in during the next sprint.

Ref https://github.com/2i2c-org/infrastructure/issues/6478


Gman0909 (approved):

Gman0909:
This has my full endorsement

PR #1001, opened 2025-07-30
Add iteration planning checklist

Gman0909:
Added a new iteration planning checklist and updated out-of-date information

Ref https://github.com/2i2c-org/meta/issues/2389

### DIFF
--- a/engineering/workflow/default-workflow.md
+++ b/engineering/workflow/default-workflow.md
@@ -77,7 +77,7 @@ Attendees of the Refinement meeting should include representatives from the foll
 - Engineering: for technical decision-making
 - Product: for product decision-making
 
-Product Managers at the meeting should ensure to come prepared to represent Business Development interests.
+Product Managers at the meeting should come prepared to represent and advocate for Business Development interests.
 
 (meetings:sprint-planning)=
 
--- a/engineering/workflow/default-workflow.md
+++ b/engineering/workflow/default-workflow.md
@@ -369,7 +369,7 @@ The Sprint Board is broken down into different columns that represent the team's
 The current queues of work represented by the board are:
 
 - {guilabel}`Upcoming P&S Initiatives` represents high level initiatives waiting to be picked up, with highest priority towards the top.
-- {guilabel}`P&S Initiatives in flight` represents high level initiatives that are actovely being worked on, with highest priority towards the top.
+- {guilabel}`P&S Initiatives in flight` represents high level initiatives that are actively being worked on, with highest priority towards the top.
 - {guilabel}`Refined` represents prioritized tasks ready to be worked, with the highest priority towards the top.
 - {guilabel}`Up Next` represents prioritized tasks that will be included into the upcoming sprint(s) with the highest priority towards the top.
 - {guilabel}`Committed` represents tasks we've committed to complete during the current sprint in the most recent sprint planning meeting. Each item should have at least one owner.
--- a/engineering/workflow/default-workflow.md
+++ b/engineering/workflow/default-workflow.md
@@ -60,29 +60,148 @@ Last day of Iteration
 :::
 
 (meetings:backlog-refinement)=
+
 ### 1. Pre-planning and refinement
 
-The Refinement meetings prepares the Iteration board for the Iteration planning
-meetings, which relies on having sequenced and prioritized tasks to be available in the board's Refined column.
+Tasks in the backlog are continually added and refined asynchronously by the whole team, ready for Iteration planning meetings, which rely on having sequenced and prioritized tasks to be available in the board's Refined column.
+The backlog is regularly culled of tasks that are stale, in a monthly Backlog refinement meeting 
 
-Backlog refinement meetings should drive:
-- creation of tasks into the sprint board's backlog column
-- promotion of tasks into the sprint board's refined column
-- prioritization of task in the sprint board's refined column
+Backlog refinement meetings should aim to
+- Identify and remove tasks that are stale or obsolete
+- Identify and action tasks that are in the wrong status or priority on the board
+- Identify tasks that need further asynchronous refinement
 
-Tasks should tie back to initiatives, and the refined column should also function as a buffer of uncommitted work to pull from if needed.
+Tasks should tie back to initiatives or contract deliverables, and the refined column should also function as a buffer of uncommitted work to pull from if needed.
 
 Attendees of the Refinement meeting should include representatives from the following teams:
 - Engineering: for technical decision-making
 - Product: for product decision-making
-- Partnerships and Community Success: for community/grant decision-making.
+
+Product Managers at the meeting should ensure to come prepared to represent Business Development interests.
 
 (meetings:sprint-planning)=
-### 2. Iteration Planning
 
-The team conducts a synchronous iteration planning meeting for 60 minutes at the beginning of each iteration.
+### 2. Iteration Planning Guide
 
-The goal of this meeting is to synchronize with one another and commit to work to do during the iteration that gets us closer to achieving the initiatives objectives.
+#### 🗓️ Between Refinement and Planning
+
+A Product Manager will:
+
+- **Add deadlines to issues that need them**  
+  A Product Manager will ensure that any issue with a contractual obligation or expected delivery has a clear due date.
+
+- **Mark contract deliverables with dates**  
+  Use labels to easily identify Contract Deliverables.
+
+- **Assign preferred sprint-end dates**  
+  If a card is intended for the upcoming sprint, let a Product Manager know and they will mark it with this sprint’s end date.
+
+- **Organize in 'Up Next'**  
+  A Product Manager manages and sorts cards in the ‘Up Next’ column for visibility.
+
+The Team will:
+
+- **Review open-source contributions**  
+  Include relevant contributions and initiatives in planning discussions.
+
+- **Advocate for priority cards**  
+  Team members should speak up for issues they’re most invested in or want to lead. Reach out to a Product Manager in the days leading up to a sprint to discuss the priority of the issues you wish to champion:
+  - First check that the issues you want to suggest are not already in the Up Next column and dated for the upcoming sprint. 
+  - Send a message to the PM at least one day prior to the planning session with a list of issues you feel need to be included in the upcoming sprint.
+  - Please include as much context as needed for the PM to understand why these issues need to be in the upcoming sprint
+  - The PM will take your suggestions under advisement and make a best effort to include them, but cannot guarantee that to be the case as other priorities and capacity will need to be considered. 
+
+#### 🗺️ Roadmap & Initiative Review
+
+The Product Manager will:
+
+- **Review roadmap tab for issues with dates**  
+  - Pull in overdue items.
+  - Reassign unrealistic dates to achievable ones.
+  - Ensure no items lag behind without follow-up.
+ 
+#### ✅ Preparing for Planning
+
+The Team will:
+
+- **Ensure issues are current**  
+  - All team members should update and clean up any issues to which they are assigned prior to the planning session
+  - Ensure the use of the Standard Definition of Done available in the General Issue template where possible.
+
+The Engineering Manager will:
+
+- **Report on last sprint’s progress**  
+  Add totals for:
+  - ✅ Done  
+  - 📂 Open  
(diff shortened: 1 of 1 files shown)


engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
 :::
 
 (meetings:backlog-refinement)=
+
 ### 1. Pre-planning and refinement
 
-The Refinement meetings prepares the Iteration board for the Iteration planning
-meetings, which relies on having sequenced and prioritized tasks to be available in the board's Refined column.
+Tasks in the backlog are continually added and refined asynchronously by the whole team, ready for Iteration planning meetings, which rely on having sequenced and prioritized tasks to be available in the board's Refined column.
```
yuvipanda, line 66:
```suggestion
Tasks in the backlog are continually added and refined asynchronously by the whole team, ready for Iteration planning meetings, which rely on having sequenced and prioritized tasks to be available in the board's Up next column.
```

Up next, right? rather than refined?

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
 
-Tasks should tie back to initiatives, and the refined column should also function as a buffer of uncommitted work to pull from if needed.
+Tasks should tie back to initiatives or contract deliverables, and the refined column should also function as a buffer of uncommitted work to pull from if needed.
 
 Attendees of the Refinement meeting should include representatives from the following teams:
 - Engineering: for technical decision-making
 - Product: for product decision-making
-- Partnerships and Community Success: for community/grant decision-making.
+
+Product Managers at the meeting should ensure to come prepared to represent Business Development interests.
```
yuvipanda, line 80:
```suggestion
Product Managers at the meeting should ensure to come prepared to represent and advocate for Business Development interests.
```

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+- **Assign preferred sprint-end dates**  
+  If a card is intended for the upcoming sprint, let a Product Manager know and they will mark it with this sprint’s end date.
+
+- **Organize in 'Up Next'**  
+  A Product Manager manages and sorts cards in the ‘Up Next’ column for visibility.
+
+The Team will:
+
+- **Review open-source contributions**  
+  Include relevant contributions and initiatives in planning discussions.
```
yuvipanda, line 105: Can we find a specific role to lead this? Our upstream fixes have vanished for a few months now.

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+
+- **Advocate for priority cards**  
+  Team members should speak up for issues they’re most invested in or want to lead. Reach out to a Product Manager in the days leading up to a sprint to discuss the priority of the issues you wish to champion:
+  - First check that the issues you want to suggest are not already in the Up Next column and dated for the upcoming sprint. 
+  - Send a message to the PM at least one day prior to the planning session with a list of issues you feel need to be included in the upcoming sprint.
+  - Please include as much context as needed for the PM to understand why these issues need to be in the upcoming sprint
+  - The PM will take your suggestions under advisement and make a best effort to include them, but cannot guarantee that to be the case as other priorities and capacity will need to be considered. 
+
+#### 🗺️ Roadmap & Initiative Review
```
yuvipanda, line 114: This too must happen before the meeting right? In that case I *think* the heading is one level off.

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+
+The Product Manager will:
+
+- **Review roadmap tab for issues with dates**  
+  - Pull in overdue items.
+  - Reassign unrealistic dates to achievable ones.
+  - Ensure no items lag behind without follow-up.
+ 
+#### ✅ Preparing for Planning
```
yuvipanda, line 123: Same comment as above for heading. Or we can explicitly state that we expect these to happen before the start of the sprint planning, either is ok.

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+
+The Engineering Manager will:
+
+- **Report on last sprint’s progress**  
+  Add totals for:
+  - ✅ Done  
+  - 📂 Open  
+  - ❌ Unplanned  
+  to the previous sprint’s **Status Update**.
+
```
yuvipanda, line 139: I would like us to add here that 'The product manager who did these things will announce that it's done in slack channel X' or similar, as a marker.

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  - Compare *actual velocity* vs. target
+  - Document the *status update*
+  - Identify questions or blockers on leftover cards
+  - Incorporate *retro feedback* to adjust workflows
+
+#### 🎯 Define Sprint Goals
+
+The Engineering Manager will:
+
+- **Establish sprint themes**
```
yuvipanda, line 156: Let's add a time limit here?

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  - 📂 Open  
+  - ❌ Unplanned  
+  to the previous sprint’s **Status Update**.
+
+#### 📆 At the Planning Session 
+
+The Engineering Manager will:
 
-The goal of this meeting is to synchronize with one another and commit to work to do during the iteration that gets us closer to achieving the initiatives objectives.
+#### 🔁 Sprint Review & Retrospective
```
yuvipanda, line 144: Let's add a time limit here?

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+
+The Engineering Manager will:
+
+- **Establish sprint themes**
+  - Ask the Head of Product and the Technical Lead to share focus areas or big goals
+  - Invite input from the whole team for broader context
+
+The Team will:
+
+- **Identify a “1” (low-effort card)**
```
yuvipanda, line 162: Let's add a time limit here?

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  - Invite input from the whole team for broader context
+
+The Team will:
+
+- **Identify a “1” (low-effort card)**  
+  Start with the easiest, quickest win for the sprint.
+
+#### 🃏 Committing Cards
+
+The Team will:
```
yuvipanda, line 167: This should be a more specific role

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  - 🧠 Estimate effort using Planning Poker or expert judgment  
+  - 🙋 Assign responsible team member(s)  
+  - 🔖 Move card to **Committed** and tag with sprint iteration
+
+- **Prioritize initiative-linked and Contract Deliverable tasks**  
+  - Reinforce the importance of completing tasks tied to Initiatives in Flight and Contract Deliverables.
+
+- **Ensure coverage of key categories**  
+  Explicitly address:
+  - Contract Deliverables
```
yuvipanda, line 184: Let's turn this into a numbered list, and explicitly state this is the priority order

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  Adjust as needed to fill capacity without overloading.
+
+- **Balance workloads**  
+  - Review assignments for each member  
+  - Shuffle cards to prevent burnout or neglect  
+  - Add or remove cards to ensure smooth sprint flow
+
+#### 📊 Final Sprint Setup
+
+- Ping anyone who was not able to attend the meeting and ensure they are aware of the tasks they have been assigned.
```
yuvipanda, line 200: Let's assign a role to this?

engineering/workflow/default-workflow.md
choldgraf, line 66: Is anybody allowed to define prioritization? AKA is anybody encouraged to move any card to any ordering in the board? Are there situations where this needs to be done by a specific person? (e.g. the product manager has the final say if there are disagreements on priority?)

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
 :::
 
 (meetings:backlog-refinement)=
+
 ### 1. Pre-planning and refinement
 
-The Refinement meetings prepares the Iteration board for the Iteration planning
-meetings, which relies on having sequenced and prioritized tasks to be available in the board's Refined column.
+Tasks in the backlog are continually added and refined asynchronously by the whole team, ready for Iteration planning meetings, which rely on having sequenced and prioritized tasks to be available in the board's Refined column.
+The backlog is regularly culled of tasks that are stale, in a monthly Backlog refinement meeting
```
choldgraf, line 67: Could you help me understand the difference between the Roadmap and the Backlog? Are they the same thing?

Based on the content here, it feels like the backlog has tasks, while the roadmap is made of P&S initiatives. Is that right? If so, that mean that refinement here is at the action/task level rather than at the level of product, is that right?

Is there a way that we can introduce refinement at the level of value, rather than at the level of task? For example, take a month-long product initiative, and spin out 3 user stories, each of which can be delivered independently of one another, in 2-3 days each. That would allow us to complete cycles of value delivery and feedback much more quickly.

engineering/workflow/default-workflow.md
choldgraf: I edited this to remove the `ensure to`. I think we should try to use as few words as possible to make this information easier to quickly digest

engineering/workflow/default-workflow.md
```diff
@@ -60,29 +60,148 @@ Last day of Iteration
...
+  - Reassign unrealistic dates to achievable ones.
+  - Ensure no items lag behind without follow-up.
+ 
+#### ✅ Preparing for Planning
+
+The Team will:
+
+- **Ensure issues are current**  
+  - All team members should update and clean up any issues to which they are assigned prior to the planning session
+  - Ensure the use of the Standard Definition of Done available in the General Issue template where possible.
```
choldgraf, line 129: I'd link to your source of truth for `Standard Definition of Done` so that somebody can quickly click-through to it.

engineering/workflow/default-workflow.md
```diff
@@ -251,8 +368,10 @@ The Sprint Board is broken down into different columns that represent the team's
 
 The current queues of work represented by the board are:
 
-- {guilabel}`Backlog` represents tasks pending refinement as driven by backlog refinement meetings.
+- {guilabel}`Upcoming P&S Initiatives` represents high level initiatives waiting to be picked up, with highest priority towards the top.
+- {guilabel}`P&S Initiatives in flight` represents high level initiatives that are actovely being worked on, with highest priority towards the top.
```
choldgraf, line 372:
```suggestion
- {guilabel}`P&S Initiatives in flight` represents high level initiatives that are actively being worked on, with highest priority towards the top.
```

engineering/workflow/default-workflow.md
Gman0909: We can certainly experiment with shorter iteration cycles, though I am not convinced the scheme you propose would be realistically applicable in many cases due to the complexities involved in getting some of our initiatives off the ground. Something to aim for aspirationally, for sure.

engineering/workflow/default-workflow.md
Gman0909: It's defined below.

engineering/workflow/default-workflow.md
Gman0909: I don't think we need to. Being too prescriptive here with the format of the meeting will only add complication and opportunity for error - I feel it's the role of an engineering manager to keep time and keep the meeting forward, and we don't need to have every minute accounted for.

engineering/workflow/default-workflow.md
Gman0909: See above

engineering/workflow/default-workflow.md
Gman0909: See above

PR #1004, opened 2025-08-05
Add support tickets to list of things to watch for in sprint planning

yuvipanda:


engineering/workflow/default-workflow.md
```diff
@@ -183,9 +183,10 @@ The Team will:
 Explicitly address:
   1. Contract Deliverables
   2. Incident Mitigation issues
-  3. Platform Initiatives  
-  4. Service Initiatives  
-before moving on to anything else. 
+  3. Support Tickets
```
aprilmj, line 186:
```suggestion
  3. Support Tickets (including workshop requests and support issues that can't be resolved in 30 minutes)  
```
To clarify that we prioritize these timely categories of support over things that might be feature requests or general feedback. Reject this suggestion if it feels like too much, though.

PR #1008, opened 2025-08-26
Set Priority field in PagerDuty to P1 for user facing incidents

yuvipanda:
Per https://2i2c.slack.com/archives/C09CGG4PTL0/p1756116481540739?thread_ts=1756116406.148179&cid=C09CGG4PTL0 and other work from @GeorgianaElena, we now use the "Priority" field in PagerDuty to specify that an incident is considered an "Outage". This lets us use PagerDuty analytics (updated once a day) to figure out how many outages we are having.

See https://2i2c-org.pagerduty.com/analytics/overview-dashboard/NjW0L-tdZdvnQxz6BjhsCQ for the outages analytics

### DIFF
--- a/projects/managed-hubs/incidents.md
+++ b/projects/managed-hubs/incidents.md
@@ -64,8 +64,8 @@ We do not have a limit on the support time we provide related to incidents (as
 opposed to Change and Guidance requests, which have a {term}`Support Budget`).
 
 We use the "Severity" field in Pagerduty to count incidents. If an incident affects end users,
-please *manually* set the Severity field to "P1" in pagerduty. This allows us to gain metrics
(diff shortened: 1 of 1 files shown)


yuvipanda:
Absolutely no idea why the build fails. I'm going to ignore it completely

choldgraf:
The reason the docs build is failing is likely because this cross-reference to an external Sphinx section no longer works:

CleanShot 2025-08-27 at 11 15 46@2x

We have a broken links issue to track these here: https://github.com/2i2c-org/team-compass/issues/500

So I suggest that we remove the documentation linkcheck as part of PR CI/CD and use that issue so that this doesn't become an "expected failure"

GeorgianaElena:
> So I suggest that we remove the documentation linkcheck as part of PR CI/CD and use that issue so that this doesn't become an "expected failure"

Makes sense @choldgraf! Expected failures are just an issue waiting to happen.

PR #1010, opened 2025-08-26
Have tech lead review incidents in last sprint

yuvipanda:
This ensures that:

1. The team is aware of the outages we have had this sprint (and the progress we are making)
2. We try to finish the incident reports before the start of next sprint, and if we don't, we flag that as an ongoing issue to be prioritized

I've also intentionally set a max time limit so this doesn't drag on. It's important to respect people's time and finish meetings on time.


choldgraf (approved):
Please just merge these PRs @yuvipanda , no need to wait for review from somebody else

PR #1016, opened 2025-09-08
Note DNS service we use

choldgraf:
We have a very sub-optimal use of domain management services right now, which is that we're split across multiple and some are owned by individuals, but this at least documents the reality!


haroldcampbell:
Thanks for the update Chris. I was scratching my head on this.

PR #1041, opened 2025-11-26
Add menu and user archetypes to business strategy 

colliand:
This moves and updates our content on buyer archetypes and membership tiers into the team compass.


choldgraf:
I've made a few edits here and will now merge this in:

- I've updated the PR title so it is more precise (generic titles like "update strategy" are hard to recognize for what they are)
- similarly I've added a brief description to the PR body.
- I've added some context that the Google doc is still the source of truth for anything that is missing here.
- cleaned up a little bit of formatting and capitalizing

PR #1045, opened 2025-12-10
Reorganize 'Communicate changes to communities' in TC

jmunroe:


jmunroe:
Related to https://github.com/2i2c-org/meta/issues/2787

PR #1098, opened 2026-04-01
Adjusts invoicing instructions for contractors

aprilmj:
Adds the new "Getting paid by CS&S" pdf link (replaces a dead link to their website. Also removes outdated information about grant codes.

Comment for reviewers: this PR represents a small tweak to our docs, but I'd like to hear if anyone who submits invoices for your time has any questions that the Compass doesn't currently answer.

### DIFF
--- a/administration/invoices.md
+++ b/administration/invoices.md
@@ -6,7 +6,7 @@ This includes any team members that are paid off of fixed-term contracts and sub
 
 ## Invoicing process
 
-1. **Read the CS&S invoicing instructions**. CS&S [has instructions to invoice here](https://www.codeforsociety.org/resources/getting-paid-by-css).
+1. **Read the CS&S invoicing instructions**. CS&S periodically updates and sends [instructions to invoice here]([https://www.codeforsociety.org/resources/getting-paid-by-css](https://drive.google.com/file/d/1_nLXaEceI1Y9dhfqZvRHhJUvzMioTm5A/view?usp=drive_link).
    This is what you'll follow.
 2. **Prepare an invoice**. If you do not have one, [here is a template to get you started](https://docs.google.com/document/d/17aTwJkmYFXwqHa2QjYsy81hEXq64yfmo5g1SCGE6aK4/edit?usp=sharing).
 3. **Confirm your grant code**. You will need to supply a **grant code** for your invoice.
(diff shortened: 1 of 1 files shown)


jnywong (approved):

GeorgianaElena (commented):
Because I had to interact with CSS recently about my invoices, I think it would have been useful for me to know:

## what fields are they expecting in the invoice
At some point they mentioned the fact that me not having a **"Due date"** field in my invoice, was generating more extra work for them. 
I think this is because Ramp automatically extracts some data from the invoice and populates some fields, that CSS is later using for scheduling, admin, etc.

It would be useful to know what are the actual fields that would make their lives easier and our invoices more likely to be paid on time.

### when is the day that they are making the payments

I believe that they have specific dates/periods when they are reviewing and initiating payments. It would be useful for me to know when these dates are so that I don't submit my invoice late, and they I have to wait out for that particular period to come again.

Thank you for opening this @aprilmj

jnywong:
Ah that's a good point, i didn't have any issues with due date, but with my most recent invoice that I submitted for March I was recommended this format for line items

| Date | Intacct Category | Grant Code | Amount |
|--|--|--|--|
| 2026-03-01 to 2026-03-31 | 5501 | 02xx 2i2c: Grant Name | $XXXX |

What wasn't helpful was that I was told to use a new grant code and i provided that without the grant name, which was rejected.

choldgraf (approved):
This looks good to me - I like linking to CS&S instructions rather than defining our own. We might also note that the cadence of "approvals" on our side is every Thursday. Our executive assistant approves both invoices and Ramp reimbursement requests in Ramp every Thursday.

aprilmj:
Good to know that Thursday is "review and approval day" - thanks, @choldgraf. 

@GeorgianaElena I started out marking my 2i2c invoices as due in 15 days, which is what I do for all other clients, and that seems to work well. I'm almost always paid in exactly 13-15 days (earlier if 15 days falls on a weekend). 

Do you all separately submit reimbursements via Ramp? I batch my receipts with my monthly invoice, and the one thing I consistently forget is that Ramp makes it really difficult for CS&S staff to read attachments; you have to print everything to a single PDF.

administration/invoices.md
```diff
@@ -30,14 +30,6 @@ Most reimbursements require a **grant code**.
 Each team member is paid from one or more grants so this helps CS&S know which account to draw from.
 The {role}`Executive Director` is the final say on which grant code is paying a given team member, but generally it is the same grant paying you over time.
 
-To find a list of grant codes:
+As new grants come in, CS&S and our team work together to allocate team members' time across grants. You will receive instructions from CS&S when you first start sending invoices and at any point when grant allocations need to change.
```
aprilmj, line 33:
```suggestion
As new grants come in, CS&S and our team work together to allocate team members' time across grants. You will receive instructions from CS&S when you first start sending invoices and at any point when grant allocations need to change.

Please use both the grant code and grant name on your invoices. CS&S will typically give you both those pieces of information.
```

administration/invoices.md
```diff
@@ -30,14 +30,6 @@ Most reimbursements require a **grant code**.
...
```
aprilmj, line 35:
```suggestion
## If an invoice is not paid when expected

If you have any difficulty in getting paid, reach out to the 2i2c People Lead or Chief of Staff. 
```

PR #1108, opened 2026-05-06
Revise communication process for infrastructure changes

choldgraf:
This revises our "communicating changes" process so we no longer export a CSV and run manual pandas/helper-sheet steps. Instead, we send the e-mail directly through HubSpot, reusing the newly-documented [Send e-mails via HubSpot](https://compass.2i2c.org/business-development/crm/#hubspot:email) process added in #1118.

The change:

- Removes the CSV export, the helper Google Sheet step, and the FreshDesk Proactive Outreach steps.
- Points the process at HubSpot, sending to our active community representatives.
- Keeps the definition of "active communities" so it's clear who the audience is.

### Follow-ups

- [ ] Confirm/create the HubSpot segment for active community representatives so the audience can be selected directly when sending.


services/interactive-computing/communicating-changes.md
```diff
@@ -1,16 +1,40 @@
...
+[^note]: This may change if we track active communities at the _deal_ level, but as of May 6 2026 this is how we define active communities.
+
+**Next, extract the e-mails with our helper sheet**
+
+1. Go to our [contacts helper google sheet](https://docs.google.com/spreadsheets/d/1XXMcDljWZIg08nHK-QU2ek0kiPnVZXpPJPil9CFl7Jo/edit?gid=641195895#gid=641195895). This is a convenience sheet to extract a unique list of contact e-mails from this export.
+2. Follow the instructions on the first tab of that sheet.
+3. **Option 1**: Paste the resulting e-mails into your e-mail client of choice and send!
+4. **Option 2**: Download the e-mails to a CSV file and follow the FreshDesk steps below.
+
+**Last, send an email using the FreshDesk template**
```
jnywong:
```suggestion
```

Duplication of 4. just above

jnywong (approved):
This is a great improvement on this process -- last time i did this it was painful so looking forward to using HubSpot rather than FreshDesk as a source of truth for email contacts

aprilmj (approved):
Approving this as a better solution for now, with the caveat that we can probably move the whole process to Hubspot.

choldgraf:
### Suggested change: Use HubSpot marketing e-mails instead of FreshDesk for communicating service changes

I looked into this a bit more, and want to make a suggestion:

- Use HubSpot e-mails instead of FreshDesk for doing "Service update" style e-mails to our communities.
- Use `Segments` to group contacts by role, consortium, etc.
- Use HubSpot's UI to send an e-mail to each segment as we wish.

This will involve using a "Drag and drop" style content UI, but we can style that to look minimal. It will have things like `Unsubscribe` at the bottom, because it's technically part of a marketing chain, but it should be much faster to send e-mails than the multi-step process we have right now.

aprilmj:
> ### Suggested change: Use HubSpot marketing e-mails instead of FreshDesk for communicating service changes

I would recommend we use what you've written here, for now, until we set up FD as an email domain. 

It's very important for the emails to come from a recognizable 2i2c address, and equally important for communities to be able to reply in a way that reaches our help desk (support@2i2c.org).

services/interactive-computing/communicating-changes.md
choldgraf: Good catch - this whole FreshDesk template section has been removed in the rewrite, so the duplication is gone. The process now sends directly through HubSpot!

choldgraf:
@aprilmj I've reworked this to move the whole process to HubSpot per your caveat, now that we merged #1118 

- **Recognizable sender:** we've now set it up to send from hello@, and I documented that!
- **Replies reaching FreshDesk:** I documented that too - also ref: #1117

choldgraf:
I'm gonna merge this one so that we don't accidentally lose track of it again!

PR #1109, opened 2026-05-13
Add cloud cost invoicing instructions to contracts

haroldcampbell:
Added section on invoicing communities for cloud costs with links to relevant documents.


choldgraf:
I've added a few cross-references, and along the way found a page that had multiple H1 headers so I fixed that too, thanks @haroldcampbell

PR #1110, opened 2026-05-15
Clean up the CRM and BD documentation

choldgraf:
This does a significant clean-up of stale, duplicated, and incorrect BD and CRM documentation. Here's a brief summary:

- Remove as much redundant and stale information as possible from BD
- Link to the in-progress hubspot guide, and use that as a source of truth
- Remove overly wordy or unnecessary paragraphs, and refer to sources of truth wherever we can
- Clean a bunch of formatting things like using ALL CAPS for headers and random spaces from copy-pasting

---

- closes https://github.com/2i2c-org/meta/issues/3158


choldgraf:
There is still a lot more to improve here, but I am going to check-point this here, and merge it in. We need to keep iterating on these docs.

PR #1128, opened 2026-06-30
Add additional BD process details

haroldcampbell:


haroldcampbell:
closes https://github.com/2i2c-org/meta/issues/3288, 2i2c-org/meta/issues/3245

business-development/guide-to-hubspot.md
```diff
@@ -0,0 +1,694 @@
...
+1. Create a related issue on the [BD requests to P&S](https://github.com/orgs/2i2c-org/projects/60) board if there is delivery work that is needed from the P&S team. These issues should be created using the `BD requests to P&S` github issue template.
+2. Ensure that a related issue has been created in the Asana [Enagements](https://app.asana.com/0/portfolio/1211914502801258/1211914511709245) board.
+3. Meet with P&S to agree when [community on-boarding](/services/interactive-computing/onboard.md) needs to be conducted.
+
+### Fields / properties
+
+:::{table} HubSpot Deal properties
+:label: tbl:areas-html
+
+<table>
```
choldgraf, line 470: we should not be putting hand-written HTML in our docs. at the very most, we should use structured markdown tables. But those are _really_ hard to maintain, so a `list-table` directive is better.

choldgraf (commented):
One quick comment, but I haven't done a full review. Before I get there, I want to first make sure this is worth our time:

Why are we converting the Google Doc into a markdown document in our compass? My rationale for putting this content in a Google Doc is that it's much easier to quickly edit a google doc, and this content is still quite fluid. At the least, I was planning to wait until the content and process have settled down before we hard-coded them in the compass.

So I want to make sure that moving all of this to markdown is actually worth the effort - can you give your rationale for why we're doing this now?

business-development/guide-to-hubspot.md
haroldcampbell: My thinking is that the google doc was temporary because things were in-flight and changing frequently. I feel that things are more ironed out now so we should be moving from the google doc to the team compass. 

Do you have a different perspective?

haroldcampbell: Also, this page is mostly a lift of google doc with minor updates.

haroldcampbell:
I haven't pulled over the Finance bit as yet because I feel we haven't finalized the API and interface for the microsite.

business-development/guide-to-hubspot.md
choldgraf: if you think that it's truly solidified, then it's OK for us to move it over, but personally I don't think any of this is solidified until:

- Everybody on the BD team knows about it and is aligned on these practices
- We've reliably followed it for at least a few months

There are also a lot of "missing data" fields in the google doc.

When we _do_ move it to markdown, we should remove as many machine content as possible. E.g., do not embed raw HTML in the compass, use a {list-table} directive instead, things like that. Otherwise, once we put things in the compass, nobody will ever update them because it's too much of a pain

haroldcampbell:
This was replaced by https://github.com/2i2c-org/team-compass/pull/1135

PR #1137, opened 2026-07-26
Convert compass to mystmd

choldgraf:
This converts our team compass to mystmd. I had to update and change a bunch of stuff along the way, because we'd done some maybe too-clever things with Sphinx already, and some of that functionality didn't exist in mystmd. Here's a quick recap of the stuff that came to mind:

- Re-worked our `{toctree}` directives into the `toc.yml` file (I kept the directives via a `{tableofcontents}` that is just for display purposes now)
- Removed some old customization that was sphinx-specific (e.g. some old CSS rules)
- Added the necessary myst config and updated all our scripts and docs to mention myst instead of sphinx
- Updated a bunch of broken links and incorrectly-used myst labels

There also seemed to be a few sections of docs that we didn't follow at all anymore (e.g. some airtable stuff and some stuff about weekly delivery workflows). I went ahead and removed them while I was at it. I'm sure there's more outdated stuff but this was just what I came across.

The actual build steps aren't that different, still just:

```
nox -s docs-live
```

This is going to touch a bunch of our files, so unless somebody objects I'm going to merge this in a day or so, and then we can clean up the text from there.

### Things that could be better but maybe worth doing later?

- The landing page is now a bit awkward, since there are no more toctrees in it...maybe just simplify it?
- The structure of the docs in general feels a bit off - we have some major and nested sections that don't feel right. The "services" section in general feels weird. But maybe that's a follow up?


choldgraf:
todo: I think there are a bunch of links that we need to add a `#` to, some other build warnings and failures!

edit: ok this is mostly done

agoose77:
I've added some site options from the Infrastructure guide. We'll want to lift that out at some point into a commmon `myst.yml`

choldgraf:
Ya I was thinking the compass could be the source of truth for site options? Feel free to push commits and or overwrite anything here if you like. Some of this was done Mid-Atlantic on little sleep haha

agoose77:
I'm going to merge this — I took a brief skim over the site, nothing exhaustive, but we'll find things as we go.

